PR #2, opened 2026-02-05
feat(cli): adding examples

innazh:
Adding examples in "after_long_help" clause to be parsed by the script in the /docs

Examples are sourced from a notion doc Ciro created: https://www.notion.so/bauplanlabs/CLI-reference-2ed9c5097c738013ae05cfb411406510

This is used to generate the cli reference page.
(will add the docs PR with the CLI stuff here later)

UPD, styling:
Screenshot 2026-02-24 at 5 02 23 PM

### DIFF
--- a/src/cli/branch.rs
+++ b/src/cli/branch.rs
@@ -10,7 +10,7 @@ use bauplan::{
     table::{GetTables, Table},
 };
 use tabwriter::TabWriter;
-use yansi::Paint;
+use yansi::Paint as _;
 
 use crate::cli::{Cli, Output, api_err_kind, checkout};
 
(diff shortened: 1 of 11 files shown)


src/cli/branch.rs
```diff
@@ -43,6 +43,23 @@ pub(crate) enum BranchCommand {
 }
 
 #[derive(Debug, clap::Args)]
+#[command(after_long_help = color_print::cstr!(r#"<bold><underline>Examples</underline></bold>
```
colinmarc: Is `color_print` necessary? the clap docs seem to suggest this is supported out of the box. maybe it needs a feature.

PR #100, opened 2026-03-03
fix: handle missing context in "forbidden errors"

colinmarc:
Because of a peculiarity, the API currently is inconsistent about whether it includes "context" in some error responses. That context isn't actually always useful or part of the public API, but rust serde is picky about it being present - at least the way that we structured the response structs thus far.

This changes the ApiErrorKind types for the "forbidden" error types to drop the context, since we don't use it and since it's often not present. For example, trying to delete the `main` branch returns a 403 without any context. This is technically a breaking change but I find it very unlikely that anyone was extracting the branch name from this error type yet.

One note about unit variants: the more idiomatic rust thing would be to have `DeleteBranchForbidden` be a "unit variant", not an empty struct. However, that breaks pyo3's `derive(pyclass)`, so we'll leave it empty.

Finally, in order to get serde to parse error responses where `context` is absent, we have to switch from thinking about it as one big adjacently-tagged enum to an externally tagged enum, and add code to reshape the json so that parsing works. This is now covered by tests.

### DIFF
--- a/src/api/error.rs
+++ b/src/api/error.rs
@@ -6,7 +6,7 @@ use crate::CatalogRef;
 /// of an error response.
 #[allow(missing_docs)]
 #[derive(Debug, Clone, Deserialize)]
-#[serde(tag = "type", content = "context", rename_all = "SCREAMING_SNAKE_CASE")]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 #[cfg_attr(
     feature = "python",
     pyo3::pyclass(module = "bauplan.exceptions", from_py_object)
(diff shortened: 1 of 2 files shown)


src/api/error.rs
```diff
@@ -223,10 +196,31 @@ impl std::fmt::Display for ApiError {
...
+
+        match serde_path_to_error::deserialize(de) {
+            Ok(kind) => ApiError::ErrorResponse {
+                status,
+                kind,
+                message: raw.message,
+            },
+            Err(e) => {
+                tracing::warn!("Failed to parse API error kind: {e}");
+                ApiError::InvalidResponse(status)
```
bigluck: What happens if the server starts returning a new error type that doesn't have a 1:1 mapping in the CLI/pysdk yet?
For example, say we introduce support for renaming a table, but the user doesn't have permission for that resource. The API would return a new specific forbidden error (e.g. `RENAME_TABLE_FORBIDDEN`), but  since that variant isn't codified in the SDK yet, it would fall through to ApiError::InvalidResponse here, losing both the error type and the message.

bigluck (approved):
thanks!

PR #169, opened 2026-04-02
feat!(pysdk): replace `all_users=True` with `filter_by_current_user=False`

### LINKED ISSUE #55: pysdk: `get_jobs` `all_users` parameter isn't intuitive from the user's perspective
The pysdk currently exposes the `protobuf` field name `all_users` directly, which reads a bit backwards; `all_users=False` means "only
  my jobs". I'd lean towards renaming it to something positive like `filter_by_current_user`, so the intent is clear at a glance.

It's a breaking change but easy to document.

Separately, I'd like to discuss the default value together. My instinct would be `filter_by_current_user=True` by default, but that could be confusing when combined with `filter_by_users`
e.g. what does `get_jobs(filter_by_current_user=True,  filter_by_users=["bob"])` mean? The server should theoretically retrieve my and Bob's jobs.

colinmarc:
This keeps the default functionality the same, but inverts the parameter to be more consistent. This is a breaking change.

Fixes #55.

### DIFF
--- a/python/bauplan/__init__.pyi
+++ b/python/bauplan/__init__.pyi
@@ -915,7 +915,7 @@ class Client:
         self,
         /,
         *,
-        all_users: bool = False,
+        filter_by_current_user: bool = True,
         filter_by_ids: str | list[str] | list[Job] | None = None,
         filter_by_users: str | list[str] | None = None,
         filter_by_kinds: str | JobKind | list[str] | list[JobKind] | None = None,
@@ -936,7 +936,8 @@ class Client:
         ```
 
         Parameters:
-            all_users: Optional[bool]: Whether to list jobs from all users or only the current user.
+            filter_by_current_user: Optional[bool]: If True (the default), only return jobs
+                belonging to the current user. Mutually exclusive with filter_by_users.
             filter_by_ids: Optional[Union[str, List[str]]]: Optional, filter by job IDs.
             filter_by_users: Optional[Union[str, List[str]]]: Optional, filter by job users.
             filter_by_kinds: Optional[Union[str, JobKind, List[Union[str, JobKind]]]]: Optional, filter by job kinds.
--- a/src/python/job.rs
+++ b/src/python/job.rs
@@ -431,7 +431,8 @@ impl Client {
     /// ```
     ///
     /// Parameters:
-    ///     all_users: Optional[bool]: Whether to list jobs from all users or only the current user.
+    ///     filter_by_current_user: Optional[bool]: If True (the default), only return jobs
+    ///         belonging to the current user. Mutually exclusive with filter_by_users.
     ///     filter_by_ids: Optional[Union[str, List[str]]]: Optional, filter by job IDs.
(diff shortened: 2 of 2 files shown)


src/python/job.rs
```diff
@@ -427,7 +427,8 @@ impl Client {
     /// ```
     ///
     /// Parameters:
-    ///     all_users: Optional[bool]: Whether to list jobs from all users or only the current user.
+    ///     filter_by_current_user: Optional[bool]: If true (the default), only return jobs
```
bigluck: nit: `True`?

bigluck (approved):
thanks, great work! just 3 minor things:
 
- do you think this change might have implications on the cli? we currently have `--all-users`, I think it's fine, but I'd like you to take a look as well
- can we add an integration test to validate that `filter_by_current_user` works?
- i think we should document that `filter_by_current_user` and `filter_by_users` can't be used together

colinmarc:
> can we add an integration test to validate that filter_by_current_user works?

We don't have integration tests for passing booleans.

> do you think this change might have implications on the cli? we currently have --all-users, I think it's fine, but I'd like you to take a look as well

Hm, yeah, we should change it on the CLI as well.

src/python/job.rs
```diff
@@ -477,7 +478,7 @@ impl Client {
 
         let job_ids = filter_by_ids.unwrap_or_default().0;
         let filter_users = filter_by_users.unwrap_or_default().0;
-        let all_users = all_users || !filter_users.is_empty();
+        let all_users = !filter_by_current_user && !filter_users.is_empty();
```
bigluck: It's monday, I spent the entire day yesterday formatting my Mac... and I didn’t sleep well last night, so I can’t guarantee a very clear review, but I will try to do my best given the delicacy of this flag :)

I think the `&&` here breaks the main case of this PR. With `filter_by_current_user=False` and no `filter_by_users`, we `send all_users=false` and the server returns only my own jobs.

colinmarc: Ok, I also find it extremely confusing, but I think what I have now is right.

bigluck (approved):

python/bauplan/__init__.pyi
```diff
@@ -936,7 +936,8 @@ class Client:
         ```
 
         Parameters:
-            all_users: Optional[bool]: Whether to list jobs from all users or only the current user.
+            filter_by_current_user: Optional[bool]: If True (the default), only return jobs
+                belonging to the current user. Mutually exclusive with filter_by_user.
```
bigluck: filter_by_users

src/python/job.rs
```diff
@@ -431,7 +431,8 @@ impl Client {
     /// ```
     ///
     /// Parameters:
-    ///     all_users: Optional[bool]: Whether to list jobs from all users or only the current user.
+    ///     filter_by_current_user: Optional[bool]: If True (the default), only return jobs
+    ///         belonging to the current user. Mutually exclusive with filter_by_user.
```
bigluck: filter_by_users

PR #180, opened 2026-04-07
docs: add cli ref page generation script

innazh:
## Summary 

Replace the hand-maintained `docs/pages/reference/cli.mdx` with a page generated directly from the clap command tree. The new `gen-cli-docs` bin walks `cli::Args::command()` and emits MDX, so the reference can stay up-to-date with the CLI.

## Changes

- `src/gen_cli_docs.rs` - new `gen-cli-docs` bin that walks `cli::Args::command()` and prints the reference page as MDX. Per command: heading with explicit anchor, `sh` synopsis, Subcommands index on parents, Arguments/Options as bullet lists, Examples block.
- `Cargo.toml` - register the bin target. 
- `docs/pages/reference/cli.mdx` - is now generated, so it's now ignored together with the rest of /reference folder.

## Regenerate

The script runs on every build, same as the sdk reference generating script.

To run manually: 
  ```sh
  cargo run --bin gen-cli-docs > docs/pages/reference/cli.mdx


colinmarc (commented):
I think this might be way easier to read if you pulled in `tera` and wrote it as a template. It's all basic loops. 

You could also try to generate components instead of "raw" markdown.

docs/pages/reference/cli.mdx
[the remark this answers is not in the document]
innazh: Yep, already removed.

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,247 @@
...
+    out.push(CommandDoc {
+        id: path.join("-"),
+        full_name: path.join(" "),
+        heading_prefix: "#".repeat(depth.min(6)),
+        is_top_level: depth == 2,
+        about: cmd.get_about().map(|a| a.to_string()).unwrap_or_default(),
+        aliases: cmd.get_visible_aliases().map(|s| s.to_string()).collect(),
+        synopsis: synopsis(path, &cmd_args, !option_args.is_empty()),
+        subcommands: subcommand_links,
+        positional: cmd_args.iter().map(|a| argument_doc(a)).collect(),
```
colinmarc:
```suggestion
        positional: cmd_args.iter().map(argument_doc).collect(),
```

This appears a bunch in this statement.

innazh:
> This is looking really neat. Does it deal with hidden aliases well? We have a few, for example `table drop` and `table ls`

Just added this! It's now displayed under "Aliases:".

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,247 @@
...
+    arg.get_help()
+        .map(|h| h.to_string().replace('\n', " "))
+        .unwrap_or_default()
+}
+
+fn option_doc(arg: &clap::Arg) -> OptionDoc {
+    debug_assert!(!arg.is_positional());
+    let value_name = arg.get_action().takes_values().then(|| default_value_name(arg));
+    OptionDoc {
+        long: arg.get_long().map(|l| format!("--{l}")).unwrap_or_default(),
```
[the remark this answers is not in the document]
innazh: `-b` option, doesn't have the long form of the flag. I'm changing it to Optional String though, and removing the `.unwrap_or_default()/.unwrap()` part altogether, replacing an empty one with a `-`.

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,247 @@
...
+        .filter(|s| is_bauplan_cmd(s))
+        .map(|sub| {
+            let mut sub_path = path.to_vec();
+            sub_path.push(sub.get_name().to_string());
+            SubcommandLink {
+                name: sub.get_name().to_string(),
+                id: sub_path.join("-"),
+                about: sub
+                    .get_about()
+                    .map(|a| a.to_string().lines().next().unwrap_or("").to_string())
```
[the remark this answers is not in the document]
innazh: yes, because I just wanted to grab a one-line summary for the subcommands, and if a user needs more info, they can click on it and go to the actual command to see what it is about.

docs/Justfile
```diff
@@ -4,6 +4,7 @@ install:
 generate:
     uv sync --reinstall-package bauplan
     uv run gen_sdk_reference.py
+    cargo run --quiet --manifest-path ../Cargo.toml --bin gen-cli-docs > pages/reference/cli.mdx
```
colinmarc: nit: this doesn't depend on anything else here, and this would read easier if it were first or last

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+        options,
+        examples: cmd
+            .get_after_long_help()
+            .or_else(|| cmd.get_after_help())
+            .map(|help| {
+                let raw = help.to_string();
+                let trimmed = raw.trim_end();
+                let stripped = trimmed
+                    .strip_prefix("Examples\n")
+                    .or_else(|| trimmed.strip_prefix("Examples\r\n"))
```
colinmarc: Do we have carriage returns anywhere?

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+        depth,
+        about: cmd.get_about().map(|a| a.to_string()).unwrap_or_default(),
+        aliases: command_aliases,
+        has_options: !options.is_empty(),
+        subcommands: subcommand_links,
+        positional,
+        options,
+        examples: cmd
+            .get_after_long_help()
+            .or_else(|| cmd.get_after_help())
```
colinmarc: I think we're pretty consistent about using `after_long_help`

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+    let (cmd_args, option_args): (Vec<&clap::Arg>, Vec<&clap::Arg>) =
+        cmd_flags.into_iter().partition(|a| a.is_positional());
+
+    let subcommand_links: Vec<SubcommandLink> = cmd
+        .get_subcommands()
+        .filter(is_bauplan_cmd)
+        .map(|sub| {
+            let mut sub_path = path.to_vec();
+            sub_path.push(sub.get_name().to_string());
+            SubcommandLink {
```
colinmarc: You could split this out into a `impl From<clap::Subcommand> for SubcommandLink`, and then it becomes:

```
.map(SubcommandLink::from).collect()
```

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+fn default_value_name(arg: &clap::Arg) -> String {
+    arg.get_value_names()
+        .and_then(|n| n.first().map(|s| s.to_string()))
+        .unwrap_or_else(|| arg.get_id().to_string().to_uppercase())
+}
+
+fn type_name(arg: &clap::Arg) -> &'static str {
+    match arg.get_action() {
+        ArgAction::SetTrue | ArgAction::SetFalse => "boolean",
+        ArgAction::Count => "int",
```
colinmarc: `ArgAction::Count` counts the number of times a flag is passed (so `bauplan --foo --foo --foo`) would be 3. I'm not sure that `int` captures that (not that we use it anywhere).

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+        _ => "string",
```
colinmarc: I think this isn't always right, we have int arguments like `--limit` etc.

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+            .get_after_long_help()
+            .or_else(|| cmd.get_after_help())
+            .map(|help| {
+                let raw = help.to_string();
+                let trimmed = raw.trim_end();
+                let stripped = trimmed
+                    .strip_prefix("Examples\n")
+                    .or_else(|| trimmed.strip_prefix("Examples\r\n"))
+                    .unwrap_or(trimmed);
+                dedent(stripped).trim_matches('\n').to_string()
```
colinmarc: I think this is redundant with the `trim_end` above.

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,241 @@
...
+    out: &mut Vec<CommandDoc>,
+) -> Result<()> {
+    let cmd_flags: Vec<&clap::Arg> = cmd
+        .get_arguments()
+        .filter(|arg| is_bauplan_arg(arg) && !global_ids.contains(arg.get_id().as_str()))
+        .collect();
+    let (cmd_args, option_args): (Vec<&clap::Arg>, Vec<&clap::Arg>) =
+        cmd_flags.into_iter().partition(|a| a.is_positional());
+
+    let subcommand_links: Vec<SubcommandLink> = cmd
```
colinmarc: nit:

```suggestion
    let subcommands: Vec<SubcommandLink> = cmd
```

(This makes the `CommandDoc { .. }` statement a bit cleaner)

src/gen_cli_docs.tera
```diff
@@ -0,0 +1,88 @@
...
+
+{% if cmd.about %}
+{{ cmd.about }}
+{% endif %}
+{% if cmd.aliases | length > 0 %}
+**Aliases:** {% for a in cmd.aliases %}`{{ a }}`{% if not loop.last %}, {% endif %}{% endfor %}
+{% endif %}
+```sh type:ignore
+{{ cmd.path_parts | join(sep=" ") }}{%
+if cmd.has_options %} [OPTIONS]{% endif %}{%
```
colinmarc: You seem to be trying to get the whitespace right here, but there are special ways with tera/jinja to do that, like `{%-` (I always forget which syntax is which, but you can look it up).

In general it's worth it to try and make the template read more naturally.

src/gen_cli_docs.rs
innazh: I just double-checked, thanks! Removing.

src/gen_cli_docs.rs
innazh: How's the below for this section? I'm not sure what's the best way to handle it. This is a suggestion I got from claudio.

```
fn type_name(arg: &clap::Arg) -> &'static str {
    match arg.get_action() {
        ArgAction::SetTrue | ArgAction::SetFalse => "boolean",
        ArgAction::Count => "count",
        _ if is_type::<u64>(arg)
            || is_type::<usize>(arg)
            || is_type::<i32>(arg)
            || is_type::<i64>(arg) =>
        {
            "int"
        }
        _ if is_type::<PathBuf>(arg) => "path",
        _ => "string",
    }
}

fn is_type<T: 'static>(arg: &clap::Arg) -> bool {
    arg.get_value_parser().type_id() == std::any::TypeId::of::<T>()
}
```

src/gen_cli_docs.tera
innazh: Added.

src/gen_cli_docs.rs
innazh: Would "count" work better here? Also, if we don't foresee ever using it,I can just remove.

docs/Justfile
innazh: Moved.

src/gen_cli_docs.rs
colinmarc: That's okay, but I wonder if we need a "type" for an arg at all? I think it's pretty obvious from the examples etc.

src/gen_cli_docs.rs
colinmarc: Yeah maybe just remove.

src/gen_cli_docs.rs
colinmarc: Like imo `--detach, boolean` is kind of more confusing than helpful - it implies `--detach false` would be valid. (But I haven't looked at how you use the type)

src/gen_cli_docs.tera
```diff
@@ -0,0 +1,93 @@
...
+{% if cmd.about %}
+{{ cmd.about }}
+{% endif %}
+{% if cmd.aliases | length > 0 %}
+**Aliases:**
+{%- for a in cmd.aliases %} `{{ a }}`{% if not loop.last %}, {% endif %}{%- endfor %}
+{% endif %}
+```sh type:ignore
+{{ cmd.path_parts | join(sep=" ") }}
+{%- if cmd.has_options %} [OPTIONS]{% endif -%}
```
colinmarc: I think you can just call `is_empty` here? But I'm not 100% sure

innazh:
```
{%- if cmd.options | length > 0 %} [OPTIONS]{% endif -%}
```

works.

src/gen_cli_docs.tera
```diff
@@ -0,0 +1,98 @@
...
+    {%- for opt in cmd.options %}
+    {{ self::cli_option(opt=opt) }}
+    {%- endfor %}
+  </CliOptions>
+  {% endif %}
+  {% if cmd.examples %}
+  <CliExamples>
+
+    ```sh
+    {{ cmd.examples }}
```
colinmarc, line 92: Now that we added `$`, this is no longer a valid `sh` snippet, so I think this might break highlighting.

I know I suggested that (adding $ to the examples), but on reflection maybe it's better to just drop it.

src/gen_cli_docs.rs
```diff
@@ -0,0 +1,261 @@
...
+        .unwrap_or_default()
+}
+
+/// Prefix examples with $.
+fn prompt_examples(s: &str) -> String {
+    let mut out = Vec::new();
+    let mut previous_nonempty_was_comment = false;
+    let mut saw_nonempty = false;
+
+    for line in s.lines() {
```
colinmarc: If we do keep this, it would be a lot cleaner if we just matched lines starting with `bauplan`.

innazh: good point, got it

src/gen_cli_docs.tera
innazh: I don't think there's much benefit in having them, so I'd rather just drop it. 

I'll make the change now in a separate commit (I'll squash before merge), and re-tag you for a review.

colinmarc (approved):

PR #185, opened 2026-04-09
Harden CI/CD

colinmarc:
This introduces the [zizmor](https://github.com/zizmorcore/zizmor) tool, which lints CI/CD for supply chain attack vectors like not pinning action versions. The idea came from Astral: https://astral.sh/blog/open-source-security-at-astral

One of the major changes (in its own commit) is to use OIDC for publishing the PyPI package.


.github/workflows/publish.yaml
```diff
@@ -97,18 +105,22 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
       - name: Download artifacts
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
 
       - name: Publish wheels to PyPI
         run: uv publish 'wheels-*/*'
```
bigluck: can we also enforce `--trusted-publishing always` here? by default it uses `automatic`, I think it's ok but I prefer to be as much explicit as possibile, if possible

bigluck (approved):
amazing work, thanks a lot @colinmarc

PR #186, opened 2026-04-09
docs: clean up more em dashes and prevent them more thoroughly

colinmarc:
The existing lint would let em dashes into the python docstrings, but then fail once the docs were generated from those docstrings (down the road).


bigluck (approved):
can we extend it to all the *.rs files too?

PR #187, opened 2026-04-09
docs: Generate SDK reference from local folder

innazh:
Generate SDK reference from local folder
- Adds a new target that generates the SDK reference from the local folder instead of the installed binary.
This is useful during development and when working on docstrings.

### DIFF
--- a/docs/gen_sdk_reference.py
+++ b/docs/gen_sdk_reference.py
@@ -3,6 +3,7 @@
 import html
 import json
 import logging
+import os
 import re
 from contextlib import contextmanager
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


docs/gen_sdk_reference.py
```diff
@@ -407,7 +408,11 @@ def main() -> None:
     output_dir = Path(__file__).parent / 'pages' / 'reference'
     output_dir.mkdir(parents=True, exist_ok=True)
 
-    bauplan = griffe.load('bauplan')
+    local_path = os.getenv('BAUPLAN_LOCAL_PATH')
```
bigluck, line 411: nit: this file uses `Path` to normalize paths, maybe you could use it here too in case the string exists

bigluck (approved):

colinmarc:
It already runs against the local source if you do `just build`. Am I missing something?

colinmarc (changes requested):
This isn't necessary - uv does this for you.

innazh:
> This isn't necessary - uv does this for you.

My bad, it wasn't working but seems like it was a caching issue. Closing this down.

colinmarc:
Ok, if you tell me more maybe we can fix the caching issue :)

PR #188, opened 2026-04-09
docs(pysdk): adding more docstring descriptions

innazh:
## Improve documentation for exceptions, state package, expectations

- Added descriptions for each of the exceptions
- Added examples for each of the expectations
- Added missing descriptions to the classes and their attributes in the `state` package
- Detect bauplan links in attribute descriptions, and allow highlights with ``s

### DIFF
--- a/python/bauplan/exceptions.pyi
+++ b/python/bauplan/exceptions.pyi
@@ -274,6 +274,9 @@ class ApiErrorKind:
         def __new__(cls, /) -> ApiErrorKind.Unauthorized: ...
 
 class BauplanError(Exception):
+    """
+    Base class for all bauplan SDK exceptions.
+    """
     def __new__(cls, /, *_args) -> BauplanError: ...
 
 class BauplanHTTPError(BauplanError):
@@ -294,6 +297,9 @@ class BauplanHTTPError(BauplanError):
 
 @final
 class TableCreatePlanApplyStatusError(BauplanError):
+    """
+    Raised when a table-create plan apply job finishes in a non-success state.
+    """
     def __new__(
         cls, /, message: str, state: TableCreatePlanApplyState
     ) -> TableCreatePlanApplyStatusError: ...
@@ -303,10 +309,16 @@ class TableCreatePlanApplyStatusError(BauplanError):
     def state(self, /) -> TableCreatePlanApplyState: ...
 
 class TableCreatePlanError(BauplanError):
+    """
+    Base class for errors raised during a table-create plan workflow.
+    """
     def __new__(cls, /) -> TableCreatePlanError: ...
 
 @final
 class TableCreatePlanStatusError(TableCreatePlanError):
+    """
+    Raised when a table-create plan job finishes in a non-success state.
+    """
     def __new__(
         cls, /, message: str, state: TableCreatePlanState
     ) -> TableCreatePlanStatusError: ...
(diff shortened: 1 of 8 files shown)


python/bauplan/exceptions.pyi
```diff
@@ -316,70 +328,163 @@ class TableCreatePlanStatusError(TableCreatePlanError):
...
+    """Raised when the provided string is not a valid `bauplan.schema.Ref`."""
+
+class NotABranchRefError(InvalidRefError):
+    """Raised when the provided `bauplan.schema.Ref` is not of type `bauplan.schema.Branch` but a `bauplan.schema.Branch` was required."""
+
+class NotATagRefError(InvalidRefError):
+    """Raised when the provided `bauplan.schema.Ref` is not of type `bauplan.schema.Tag` but a `bauplan.schema.Tag` was required."""
+
+class NotAWriteBranchRefError(NotABranchRefError):
+    """Raised when the provided `bauplan.schema.Ref` is a `bauplan.schema.Branch` but not writable."""
```
bigluck: `Raised when a write operation is attempted against a ref that is not a branch (e.g. a tag or detached commit hash)`

python/bauplan/exceptions.pyi
```diff
@@ -316,70 +328,163 @@ class TableCreatePlanStatusError(TableCreatePlanError):
...
+    """Raised when a merge cannot be completed due to conflicting changes."""
+
+class NamespaceIsNotEmptyError(UpdateConflictError):
+    """Raised when attempting to delete a `bauplan.schema.Namespace` that still contains tables."""
+
+class RevertDestinationTableExistsError(UpdateConflictError):
+    """Raised when the destination of a revert operation already exists."""
+
+class RevertIdenticalTableError(UpdateConflictError):
+    """Raised when the source and destination of a revert are the same table."""
```
bigluck: nit: there's a subtle distinction between "same table" (same table UUID) and "table with the same snapshot" (same table UUID + same snapshot UUID).

The message here implies the first case, while the APIs check for the second.

For example... suppose I have the table `my_namespace.my_table_name` in the main branch with `table_uuid = t1`, and `snapshot_uuid = s1`, and in my branch I have `my_namespace.my_table_name` with the same `t1` and `s1`.  If I try to restore the table, it will fail with this error, because the 2 tables are in a identical state.

However, if in my branch I have `t1` with `s2` <--, and I try to restore it in main, it will work.

Another example: if in my branch I have a `t2` & `s21` (a completely different table) and I try to restore it in main (all these cases are with `replace=True`, of course), the revert will not fail.

python/bauplan/exceptions.pyi
```diff
@@ -316,70 +328,163 @@ class TableCreatePlanStatusError(TableCreatePlanError):
...
+    """Raised when the `bauplan.schema.Branch` head hash has changed since it was last read."""
+
+class MergeConflictError(UpdateConflictError):
+    """Raised when a merge cannot be completed due to conflicting changes."""
+
+class NamespaceIsNotEmptyError(UpdateConflictError):
+    """Raised when attempting to delete a `bauplan.schema.Namespace` that still contains tables."""
+
+class RevertDestinationTableExistsError(UpdateConflictError):
+    """Raised when the destination of a revert operation already exists."""
```
bigluck, line 457: nit: might it make sense to mention that the error occurs when `replace` is not `True`?
`Raised when the destination of a revert operation already exists and replace is not True.`

python/bauplan/exceptions.pyi
```diff
@@ -316,70 +328,163 @@ class TableCreatePlanStatusError(TableCreatePlanError):
...
+    """Raised when the provided `bauplan.schema.Ref` is not of type `bauplan.schema.Branch` but a `bauplan.schema.Branch` was required."""
+
+class NotATagRefError(InvalidRefError):
+    """Raised when the provided `bauplan.schema.Ref` is not of type `bauplan.schema.Tag` but a `bauplan.schema.Tag` was required."""
+
+class NotAWriteBranchRefError(NotABranchRefError):
+    """Raised when the provided `bauplan.schema.Ref` is a `bauplan.schema.Branch` but not writable."""
+
+class SameRefError(InvalidRefError):
+    """Raised when the source and destination `bauplan.schema.Ref` values of an operation are the same."""
```
bigluck: nit: `Raised when the source and destination `bauplan.schema.Ref` resolve to the same commit hash, making the operation a no-op.`?

python/bauplan/exceptions.pyi
```diff
@@ -316,70 +328,163 @@ class TableCreatePlanStatusError(TableCreatePlanError):
...
+    """Raised when creating a table that already exists on the target `bauplan.schema.Ref`."""
+
+class TagExistsError(UpdateConflictError):
+    """Raised when creating a `bauplan.schema.Tag` that already exists."""
+
+class NamespaceExistsError(UpdateConflictError):
+    """Raised when creating a `bauplan.schema.Namespace` that already exists on the target `bauplan.schema.Ref`."""
+
+class NamespaceUnresolvedError(ConflictError):
+    """Raised when a `bauplan.schema.Namespace` reference cannot be resolved."""
```
bigluck: nit: "cannot be resolved" is vague and could be confused with `NamespaceNotFoundError` (namespace doesn't exist). The actual scenario is a conflict between two sources of namespace information in the same request.

For example, suppose I call:

```python
client.get_table(table="my_namespace.my_table", ref="main", namespace="other_namespace")
```

The table name is fully qualified (contains a dot, so it already embeds `my_namespace`), but I'm also passing an explicit `namespace="other_namespace"`. The catalog can't decide which one to use, so it fails with `NAMESPACE_UNRESOLVED`.

If I call `client.get_table(table="my_namespace.my_table", ref="main")` it works fine. Same if I call `client.get_table(table="my_table", ref="main", namespace="my_namespace")`.

The error only fires when both are provided and they create ambiguity.

Maybe here we could say something like: `Raised when a table identifier includes a namespace (e.g. ns.table) and an explicit namespace parameter is also provided, creating an ambiguous reference.` ?

Same applies to `delete_table` and `revert_table`.

bigluck (commented):
Thanks inna, good work overall! the state docstrings and the expectation examples are solid and well-written... I left some comments with detailed explanations of when these are generated, so that in case you can find a more elegant way than mine to explain to the user :)

python/bauplan/exceptions.pyi
innazh: How's

> Raised when a write operation is attempted against a `bauplan.schema.Ref` that is not of type `BRANCH`.

python/bauplan/exceptions.pyi
innazh: What do you think about:

> """Raised by `Client.revert_table` when the destination of a revert operation already exists, and `replace` flag is not set to `True`."""

innazh: But then I'm wondering if it's a good thing in general to reference the methods that emit these exceptions?

I tried to avoid it at first because maybe we want to keep certain exceptions more general and might use them in more methods in the future but if we're sure, say, `RevertDestinationTableExistsError` is only used for this method, then I'd rather mention it. Do you have a vision you can share in regards to this?

python/bauplan/exceptions.pyi
innazh: >     "Raised when the source and destination `bauplan.schema.Ref`
>     resolve to the same commit hash, making an operation invalid."

python/bauplan/exceptions.pyi
innazh: Great point! I think saying `the same table` here implies that this will be the table with the same uuid that points to the same data (snapshot). So if we want to keep it brief, I'd leave it. Otherwise, if you'd think we benefit from being more explicit, see below.

I have a couple of options on how we can phrase this:

1:
>     "Raised by `Client.revert_table` when the source and destination of a revert 
>      are the same table, meaning both refer to the same underlying data."

2:
>     "Raised by `Client.revert_table` when the source and destination of a revert 
>      are the same table, meaning both have the same UUID, and point to the same snapshot."

Note: I raised the point of referencing a function name in another comment below, and depending on the outcome of our discussion will change the above.

python/bauplan/exceptions.pyi
innazh: How's this?

> Raised when a `bauplan.schema.Namespace` is specified both in the table name (e.g. ns.table) and through an explicit `namespace` parameter, resulting in an ambiguous reference

python/bauplan/exceptions.pyi
bigluck: 👍

python/bauplan/exceptions.pyi
bigluck: I don't know if I'd provide the context of who's generating the error, even if it's implicit that this error is generated by that call.
That said,they both seem a bit too verbose. The first one also has a semantic problem because I could create a schema evolution for a table that would generate a new snapshot with a schema evolution but without changing the state of the data in the lake.

What about something a little more specific like "the source and destination point to the same table snapshot"?

python/bauplan/exceptions.pyi
bigluck: I don't like to reference who generates an error, as I mentioned in a previous comment I don't think it gives much value, and knowing that in the future we will define the list of errors generated per endpoint, I would avoid specifying it even in the description.

bigluck: in any case the error implied by the system in this case is: you are trying to restore from a ref X in a branch Y a resource/table whose name is already used by another resource/table

python/bauplan/exceptions.pyi
bigluck: I'm not convinced by the term invalid. It's not formally invalid, you're asking to do nothing in the end (same source and same ref means you don't want to change anything); I think semantically no-op is more correct, because today we treat the no-op case as an error.

python/bauplan/exceptions.pyi
bigluck: 👍

python/bauplan/exceptions.pyi
innazh: Agreed.

python/bauplan/exceptions.pyi
innazh: If we're not mentioning what function generates the error, then we should skip mentioning the `replace` flag, since without mentioning the related operation, it might not make much sense. What do you think?

innazh: So this would stay as `Raised when the destination of a revert operation already exists.`

bigluck: > Raised when the destination of a revert operation already exists

i like it

bigluck (approved):

PR #191, opened 2026-04-13
feat!(cli): replace on/off flags with simple boolean flags

colinmarc:
This includes `--cache`, `--transaction`, and `--strict`.

(I'm teeing this up for when we're ready to ship `0.2.0`.)

### DIFF
--- a/docs/pages/reference/cli.mdx
+++ b/docs/pages/reference/cli.mdx
@@ -111,7 +111,7 @@ bauplan run [flags]
 | --- | --- | --- | --- | --- |
 | `--arg` | `-a` | string (repeatable) | - | Arguments to pass to the job. Format: key=value |
 | `--project-dir` | `-p` | string | "." | Path to the root Bauplan project directory |
-| `--cache` | - | string | - | Set the cache mode. [on, off] |
+| `--nocache` | - | string | - | Disable caching |
 | `--summary-no-trunc` | - | boolean | false | Do not truncate summary output |
 | `--preview` | - | string | - | Set the preview mode. [on, off, head, tail] |
 | `--strict` | - | string | - | Exit upon encountering runtime warnings (for example, invalid column output). [on, off] |
@@ -119,7 +119,7 @@ bauplan run [flags]
 | `--namespace` | `-n` | string | - | Namespace to run the job in. If not set, the job will be run in the default namespace for the project |
 | `--ref` | `-r` | string | - | Ref or branch name from which to run the job |
 | `--dry-run` | - | boolean | false | Dry run the job without materializing any models |
-| `--transaction` | `-t` | string | - | Run the dag as a transaction. Will create a temporary branch where models are materialized. Once all models succeed, it will be merged to branch in which this run is happening in |
+| `--no-transaction` | `-t` | string | - | Disable transactional branching. Without this flag, the server will create a temporary branch where models are materialized. Once all models succeed, it will be merged to branch in which this run is happening in |
 | `--detach` | `-d` | boolean | false | Run the job in the background instead of streaming logs |
 | `--priority` | - | int | - | Set the job priority (1-10, where 10 is highest priority) |
 | `--help` | `-h` | - | - | Show help |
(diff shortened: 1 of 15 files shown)


docs/pages/reference/cli.mdx
```diff
@@ -111,15 +111,15 @@ bauplan run [flags]
 | --- | --- | --- | --- | --- |
 | `--arg` | `-a` | string (repeatable) | - | Arguments to pass to the job. Format: key=value |
 | `--project-dir` | `-p` | string | "." | Path to the root Bauplan project directory |
-| `--cache` | - | string | - | Set the cache mode. [on, off] |
+| `--nocache` | - | string | - | Disable caching |
```
bigluck: shouldn't it be `--no-cache` here? shouldn't the column also be a `boolean` type with default to `false`?

docs/pages/reference/cli.mdx
```diff
@@ -111,15 +111,15 @@ bauplan run [flags]
...
 | `--summary-no-trunc` | - | boolean | false | Do not truncate summary output |
 | `--preview` | - | string | - | Set the preview mode. [on, off, head, tail] |
 | `--strict` | - | string | - | Exit upon encountering runtime warnings (for example, invalid column output). [on, off] |
```
bigluck: this column should also become of type `boolean`, with default to `false` (and without `[on, off]` in the description)

docs/pages/reference/cli.mdx
```diff
@@ -111,15 +111,15 @@ bauplan run [flags]
...
 | `--param` | - | string (repeatable) | - | Set a parameter for the job. Format: key=value. Can be used multiple times |
 | `--namespace` | `-n` | string | - | Namespace to run the job in. If not set, the job will be run in the default namespace for the project |
 | `--ref` | `-r` | string | - | Ref or branch name from which to run the job |
 | `--dry-run` | - | boolean | false | Dry run the job without materializing any models |
-| `--transaction` | `-t` | string | - | Run the dag as a transaction. Will create a temporary branch where models are materialized. Once all models succeed, it will be merged to branch in which this run is happening in |
+| `--no-transaction` | `-t` | string | - | Disable transactional branching. Without this flag, the server will create a temporary branch where models are materialized. Once all models succeed, it will be merged to branch in which this run is happening in |
```
bigluck: `boolean` default `false`

bigluck: Can you double-check `-t`? From what I see, support for it has been removed from the code

tests/cli/run.rs
bigluck: Now that transaction is explicitly on, we should convert this test to a `without_transaction` one, using the `--no-transaction` flag

docs/pages/reference/cli.mdx
```diff
@@ -200,7 +200,7 @@ bauplan query [flags] [SQL]
 | Flag | Short | Type | Default | Description |
 | --- | --- | --- | --- | --- |
 | `--no-trunc` | - | boolean | false | Do not truncate output |
-| `--cache` | - | string | - | Set the cache mode. [on, off] |
+| `--no-cache` | - | string | - | Disable caching |
```
bigluck: `boolean` default `false`

docs/pages/reference/cli.mdx
colinmarc: This file will be autogenerated in #180, then I can rebase.

bigluck (approved):
Ok this is an interesting breaking change.

Look at the prompt in `examples/projects/executive-questions-with-mcp/template/.github/workflows/claude.yml`: we ask the agent to run `--dry-run --strict on`.
In theory we should change it to `--strict`, but if a customer takes this example from `main` they would get an invalid instruction, because the released sdk is still 0.1.x and there `--strict` alone does not work.

Same problem for the docs: we can update them, but if someone deploys a new docs page to prod, we would expose documentation for an sdk version that is not released yet.

colinmarc:
Fixed `claude.yml`.

I think as long as we merge it around the time we do the release we don't have to worry. The published docs always reflect the latest version.

PR #197, opened 2026-04-15
feat: longbow

colinmarc:
This adds a new protocol for serving query results directly from the runtimes called "longbow", based on iroh and arrow IPC.

### DIFF
--- a/bauplan-longbow/src/server.rs
+++ b/bauplan-longbow/src/server.rs
@@ -1,0 +1,137 @@
+use arrow::{
+    array::RecordBatch,
+    datatypes::SchemaRef,
+    ipc::{
+        CompressionType,
+        writer::{IpcWriteOptions, StreamWriter},
+    },
+};
+use bytes::{BufMut, BytesMut};
+use iroh::{
+    Endpoint, SecretKey,
+    endpoint::{Connection, ConnectionState, IncomingZeroRtt, SendStream, presets::Preset},
+};
+use tracing::{debug, error};
+
+use crate::{Error, StreamToken};
+
+/// Binds an endpoint and waits for the first connection, then returns it. This
+/// will run forever and has no inherent timeout.
+async fn accept_connection(
+    preset: impl Preset,
+    secret_key: SecretKey,
+) -> Result<(Endpoint, Connection<IncomingZeroRtt>), Error> {
+    let endpoint = Endpoint::builder(preset)
+        .secret_key(secret_key)
+        .bind()
+        .await?;
+
+    loop {
+        let Some(incoming) = endpoint.accept().await else {
+            // The endpoint was dropped.
+            return Err(Error::NoPeer);
+        };
+
+        match incoming.accept() {
+            Ok(acc) => return Ok((endpoint, acc.into_0rtt())),
+            Err(err) => {
+                error!(?err, "handshake failed, listening again");
+                continue;
+            }
+        }
+    }
+}
+
+/// Accept the next bidirectional stream on a connection, reading the
+/// one-byte [`StreamToken`] the client sends to identify the stream type.
+async fn accept_stream<C: ConnectionState>(
+    conn: &Connection<C>,
+) -> Result<(StreamToken, SendStream), Error> {
+    let (send, mut recv) = conn.accept_bi().await?;
+
+    let mut token_buf = [0u8; 1];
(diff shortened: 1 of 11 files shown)


src/cli/query.rs
```diff
@@ -151,6 +152,56 @@ pub(crate) async fn handle(cli: &Cli, args: QueryArgs) -> anyhow::Result<()> {
     )
     .await?;
 
+    progress.set_message("Fetching results...");
+
+    let tp = cli.traceparent();
+    let longbow_key = if resp.longbow_public_key.is_empty() {
+        None
+    } else {
+        eprintln!("using LONGBOW");
```
bigluck: there's a debug print here

src/cli/query.rs
```diff
@@ -151,6 +152,56 @@ pub(crate) async fn handle(cli: &Cli, args: QueryArgs) -> anyhow::Result<()> {
...
+) -> anyhow::Result<(Schema, impl Stream<Item = FlightResult<RecordBatch>>)> {
+    if let Some(public_key_bytes) = longbow_public_key {
+        let public_key = bauplan_longbow::iroh::PublicKey::try_from(public_key_bytes.as_slice())
+            .context("invalid longbow public key")?;
+        let preset = BauplanPreset::default();
+        let addr = preset.add_relay_urls(bauplan_longbow::iroh::EndpointAddr::new(public_key));
+
+        let (schema, stream) = bauplan_longbow::fetch_query_results(preset, addr)
+            .await
+            .context("failed to fetch query results")?;
```
bigluck: nit: this connection does not have a timeout (we should have a client timeout, but I don't think it's used here)

bauplan-longbow/src/server.rs
```diff
@@ -0,0 +1,138 @@
...
+        Ok(Self {
+            endpoint,
+            send,
+            writer,
+        })
+    }
+
+    /// Buffers a record batch to be sent. Data can be flushed with [flush].
+    pub async fn send_record_batch(&mut self, batch: &RecordBatch) -> Result<(), Error> {
+        self.writer.write(batch)?;
```
bigluck, line 108: Sorry maybe I'm missing something, but here `send_record_batch` only writes to memory no? If the runtime call it many times before `finish()`, does the client have to wait all data is buffered before receiving the first batch? Or there is some auto-flush somewhere that I don't see?

bauplan-longbow/src/server.rs
```diff
@@ -0,0 +1,138 @@
...
+    loop {
+        let Some(incoming) = endpoint.accept().await else {
+            // The endpoint was dropped.
+            return Err(Error::NoPeer);
+        };
+
+        match incoming.accept() {
+            Ok(acc) => return Ok((endpoint, acc.into_0rtt())),
+            Err(err) => {
+                error!(?err, "handshake failed, listening again");
```
bigluck, line 38: if a peer keep failing the handshake, we loop forever, and every attempt is logged at error level. I'm afraid this can leave the server spinning forever

bigluck (approved):

bauplan-longbow/src/server.rs
colinmarc: the QUIC conn should be sending as fast as possible in the background, adhering to congestion control.

bauplan-longbow/src/server.rs
colinmarc: The intent is that the caller uses `tokio::time::timeout`.

colinmarc: (And indeed we do on the server)

bauplan-longbow/src/server.rs
bigluck: maybe I'm still missing something: the `StreamWriter` writes into a `BytesMut` via `byts::buf::writer`, so until we call `flush()` nothing is handed to `self.send.write_chnk`, no?
My understanding is that QUIC congestion control kicks in only after flush. But I don't remember if the runtime calls `flush()` after every batch, or somewhere else the buffer is drained automatically?

colinmarc: Sorry, you are totally right, that's the behavior. On reflection it's probably good to just call flush in `write_batch`, since it just goes into the quic write buffer anyway.

colinmarc: Done, I also made `flush` private.

PR #200, opened 2026-04-15
Document CLI defaults more consistently

innazh:
This PR makes CLI defaults more explicit and consistent in help output.

Changes:
- Set explicit Clap defaults for run options that already have runtime defaults:
      - `--cache`
      - `--preview`
      - `--strict`
      - `--transaction`
- Set `--project-dir` default to `.` for:
      - run
      - parameter ls
      - parameter rm
      - parameter set
- Set commit `--format` to default to medium
- Set query `--cache` to default to on

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -51,6 +51,15 @@ pub(crate) enum Output {
     Tty,
 }
 
+impl std::fmt::Display for Output {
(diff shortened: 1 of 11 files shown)


colinmarc (approved):
Great work! Thank you!

PR #203, opened 2026-04-16
docs: add utilities scripts

giacomopiccinini:
Add first two utilities to manage transactional branches. 

The first one, lets the user inspect and print the branches. This is mostly useful for e.g. agents, or to have a quick overview of how Bauplan operates. 

The second one, lets the user delete transactional branches in a definite time-window (e.g. before X, after Y, or between Y and X). A `dry-run` option allows the user to _not_ delete the branch: they are just printed.

### DIFF
--- a/examples/utils/01_find_transactional_branches/main.py
+++ b/examples/utils/01_find_transactional_branches/main.py
@@ -1,0 +1,63 @@
+import bauplan
+
+import re
+import typer
+from typing import Annotated
+from collections import Counter
+
+
+def find_transactional_branches(
+    profile: Annotated[str, typer.Option(help="Bauplan profile to use")] = "default",
+    verbose: Annotated[
+        bool, typer.Option(help="Print all transactional branches")
+    ] = False,
+) -> None:
+    """
+    Retrieve transactional branches that can be pruned to clean up the lakehouse.
+
+    A transactional branch is created for any Bauplan command altering the state of the lake and remains open if job execution fails for some reason.
+
+    Given a transactional branch, its corresponding job should be in one of two states:
+        - 'failed': branch can be pruned or inspected for debugging
+        - 'running': branch cannot be pruned, job still in flight
+
+    Prints a breakdown of transactional branches by command, sorted from most to least frequent,
+    optionally followed by the full list of branch names.
+    """
+    client = bauplan.Client(profile=profile)
+
+    # Counter for commands
+    counter = Counter()
+
+    # Buffer to be used to print out branch names in verbose mode
+    buffer = []
+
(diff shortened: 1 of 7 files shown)


examples/utils/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "bauplan-utils"
+version = "0.1.0"
+description = "Utilities to manage Bauplan"
+readme = "README.md"
+requires-python = ">=3.13"
+dependencies = [
+    "bauplan>=0.1.10",
```
bigluck: Can you use `~=0.1.10` instead of `>=0.1.10` here? 0.2 may introduce breaking changes

examples/utils/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "bauplan-utils"
+version = "0.1.0"
+description = "Utilities to manage Bauplan"
+readme = "README.md"
+requires-python = ">=3.13"
```
bigluck: I would limit the python version here to those supported by `bauplan`, since we have binary files that we need pre-compile, and `>=` might not work with `0.1.x` if python `3.15` comes out tomorrow and we don't have any `0.1.x` version with support for `3.15`

examples/utils/pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+dependencies = [
+    "bauplan>=0.1.10",
+    "typer>=0.24.1",
```
bigluck: I know we have a `uv.lock` file and we use `typer` in a very basic way, but I would also suggest limiting this dep with `~=`

examples/utils/01_find_transactional_branches/main.py
```diff
@@ -0,0 +1,63 @@
+import bauplan
```
bigluck, line 1: @colinmarc do you remember if our CI/CD also verifies these files with `ruff` and the bauplan signature checker?

bigluck (commented):
first pass, thanks

examples/utils/pyproject.toml
giacomopiccinini: So shall we stick to, say, 3.12?

colinmarc: 3.13 please (and we badly need to add 3.14 soon)

examples/utils/01_find_transactional_branches/main.py
colinmarc: No, we don't typecheck the examples or the test fixtures, because bauplan models don't really typecheck successfully today.

bigluck (approved):

PR #204, opened 2026-04-16
docs: add -function postfix to fn links

innazh:
### Add -function postfix to function entries in SDK reference
Functions and classes in the SDK reference currently share the same link-generated format, making them both have the same link for stuff like `class Model` and `def model()`. This PR appends a -function postfix to function entries so these can be clearly differentiated.

This was brought up today by @giacomopiccinini.


bigluck (approved):
Thanks!

giacomopiccinini (approved):
Nice, thanks @innazh!

PR #214, opened 2026-04-22
fix: percent-encode URL path segments

bigluck:
User-provided identifiers (branches, tags, namespaces, tables, refs) were interpolated raw into request paths, breaking values that contain reserved characters such as `/`.
Each path segment is now encoded per RFC 3986, leaving final validation of the decoded value to the server.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,8 +1,10 @@
 // ApiError is just over the threshold.
 #![allow(clippy::result_large_err)]
 
+use std::borrow::Cow;
 use std::io::Read;
 
+use percent_encoding::{AsciiSet, CONTROLS, PercentEncode, utf8_percent_encode};
 use serde::{Deserialize, Serialize};
 use tracing::warn;
 
@@ -22,6 +24,32 @@ pub(crate) mod testutil;
 
 pub use error::*;
 pub use paginate::*;
+
+/// A percent-encoded URL path for an API request.
+#[derive(Debug)]
+pub struct PathArgs(Cow<'static, str>);
+
+fn encode_segment(s: &str) -> PercentEncode<'_> {
+    // WHATWG path percent-encode set (https://url.spec.whatwg.org/#path-percent-encode-set)
+    // extended with `/` and `%` to treat the input as a single segment.
+    const SEGMENT: &AsciiSet = &CONTROLS
+        .add(b' ').add(b'"').add(b'<').add(b'>').add(b'`')
+        .add(b'#').add(b'?').add(b'{').add(b'}')
+        .add(b'/').add(b'%');
+    utf8_percent_encode(s, SEGMENT)
+}
+
+macro_rules! urlformat {
+    ($fmt:literal) => {
+        $crate::api::PathArgs(::std::borrow::Cow::Borrowed($fmt))
+    };
+    ($fmt:literal, $($arg:expr),+ $(,)?) => {
+        $crate::api::PathArgs(::std::borrow::Cow::Owned(format!(
+            $fmt, $($crate::api::encode_segment($arg)),+
+        )))
+    };
+}
+pub(crate) use urlformat;
 
 #[derive(Debug, Deserialize)]
 struct RawMetadata {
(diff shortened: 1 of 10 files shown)


src/api.rs
```diff
@@ -23,6 +24,16 @@ pub(crate) mod testutil;
 pub use error::*;
 pub use paginate::*;
 
+/// Path segment encode set (RFC 3986 §3.3).
+const SEGMENT: &AsciiSet = &CONTROLS
+    .add(b' ').add(b'"').add(b'#').add(b'<').add(b'>').add(b'?').add(b'`')
+    .add(b'{').add(b'}').add(b'[').add(b']').add(b'\\').add(b'^').add(b'|')
+    .add(b'/').add(b'%');
```
colinmarc: Why not follow this? https://url.spec.whatwg.org/#fragment-percent-encode-set

src/api.rs
```diff
@@ -23,6 +24,16 @@ pub(crate) mod testutil;
...
+
+pub(crate) fn encode_segment(s: &str) -> PercentEncode<'_> {
```
colinmarc: Instead of this, how about a `macro_rules urlformat` that passes through to `format_args` and automatically calls it on every parameter? It could return a newtype like:

```
struct PathArgs(std::fmt::FormatArgs);

// Allows using non-parameterized paths with `.into()`
impl From<&'static str> for FormattedPath { ... }
```

src/api.rs
```diff
@@ -23,6 +24,16 @@ pub(crate) mod testutil;
 pub use error::*;
 pub use paginate::*;
 
+/// Path segment encode set (RFC 3986 §3.3).
+const SEGMENT: &AsciiSet = &CONTROLS
```
colinmarc: You can move this into `encode_segment`.

src/api.rs
```diff
@@ -47,7 +65,8 @@ pub trait ApiRequest: Sized {
     /// The corresponding response type.
     type Response: ApiResponse;
 
-    /// The path that the request should take.
+    /// The path that the request should take. Any dynamic segments must be
+    /// percent-encoded; use [`urlformat!`] to guarantee this.
```
colinmarc: If we return a newtype, we can force this.

src/api.rs
```diff
@@ -183,3 +202,20 @@ impl ApiResponse for CatalogRef {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::encode_segment;
+
+    #[test]
+    fn encode_segment_covers_expected_chars() {
```
colinmarc: Maybe test urlformat here?

colinmarc (approved):

src/api.rs
```diff
@@ -23,6 +24,29 @@ pub(crate) mod testutil;
 pub use error::*;
 pub use paginate::*;
 
+/// A percent-encoded URL path for an API request.
+#[derive(Debug)]
+pub struct PathArgs(pub(crate) String);
+
+pub(crate) fn encode_segment(s: &str) -> PercentEncode<'_> {
```
colinmarc:
```suggestion
fn encode_segment(s: &str) -> PercentEncode<'_> {
```

colinmarc (approved):

src/api.rs
```diff
@@ -23,6 +24,29 @@ pub(crate) mod testutil;
 pub use error::*;
 pub use paginate::*;
 
+/// A percent-encoded URL path for an API request.
+#[derive(Debug)]
+pub struct PathArgs(pub(crate) String);
```
colinmarc:
```suggestion
pub struct PathArgs(String);
```

src/api.rs
```diff
@@ -183,3 +216,38 @@ impl ApiResponse for CatalogRef {
...
+
+#[cfg(test)]
+mod tests {
+    use std::borrow::Cow;
+
+    #[test]
+    fn urlformat_static_is_borrowed() {
+        let p = urlformat!("/catalog/v0/branches").0;
+        assert!(matches!(p, Cow::Borrowed(_)));
+        assert_eq!(p, "/catalog/v0/branches");
```
colinmarc: nit: this can me a single assert

colinmarc (approved):

PR #216, opened 2026-04-22
fix: update teardown behavior for p2p connections

colinmarc:
#197 introduced p2p connections for fetching query results, but I accidentally pushed the wrong revision and it had some experimental code that I wasn't really happy with.

### DIFF
--- a/bauplan-longbow/src/client.rs
+++ b/bauplan-longbow/src/client.rs
@@ -155,11 +155,7 @@ pub async fn fetch_query_results(
                         return Ok(Some((batch, state)));
                     }
                 }
-                None => {
-                    // The conn.close() in Drop sends the close frame.
-                    // Endpoint cleanup happens asynchronously.
-                    return Ok(None);
-                }
+                None => return Ok(None),
             }
         }
     }));
--- a/bauplan-longbow/src/server.rs
+++ b/bauplan-longbow/src/server.rs
@@ -125,11 +125,7 @@ impl ArrowIPCServer {
 
         // Signal that no more data will be sent.
         let _ = self.send.finish();
-
-        // Wait briefly for the FIN to be acknowledged, but don't block
-        // indefinitely — QUIC's reliability layer ensures delivery.
-        let _ =
-            tokio::time::timeout(std::time::Duration::from_millis(500), self.send.stopped()).await;
+        self.send.stopped().await.map_err(|_| Error::StreamClosed)?;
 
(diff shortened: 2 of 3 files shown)


bauplan-longbow/src/client.rs
```diff
@@ -155,11 +155,7 @@ pub async fn fetch_query_results(
                         return Ok(Some((batch, state)));
```
bigluck, line 155: random comment in a random file... 

> nit: this connection does not have a timeout (we should have a client timeout, but I don't think it's used here)

https://github.com/BauplanLabs/bauplan/pull/197/changes#diff-1edcd8a0e8e4044d0b9fa4765f24b64471614e071a8a4b149257b126b0b98836

you applied the fix only to the cli/query.rs file:

https://github.com/BauplanLabs/bauplan/pull/197/changes#diff-d2db564408f8bda35a94a5b01b133e4113f1a5dfcdd8b637b274a0c9aa64298dR191-R194

but the `python/query.rs` file is still using the a naked `.await`:

https://github.com/BauplanLabs/bauplan/pull/197/changes#diff-1edcd8a0e8e4044d0b9fa4765f24b64471614e071a8a4b149257b126b0b98836R124-R126

bauplan-longbow/src/server.rs
```diff
@@ -125,11 +125,7 @@ impl ArrowIPCServer {
 
         // Signal that no more data will be sent.
         let _ = self.send.finish();
-
-        // Wait briefly for the FIN to be acknowledged, but don't block
-        // indefinitely — QUIC's reliability layer ensures delivery.
-        let _ =
-            tokio::time::timeout(std::time::Duration::from_millis(500), self.send.stopped()).await;
+        self.send.stopped().await.map_err(|_| Error::StreamClosed)?;
```
bigluck, line 128: if `stopped()` return an error here, the `?` early-returns and `self.endpoint.close().await` just below is skipped, so we could lose the graceful endpoint shutdown... wondering if we should close the endpoint anyway before propagating the error

colinmarc: No, that should be fine. The drop on `Endpoint` should be enough and in that case we're in a failure mode either way.

bauplan-longbow/src/client.rs
colinmarc: Ah yeah, good catch.

bigluck (approved):
thks

PR #218, opened 2026-04-22
build: fix `release-plz pr`

colinmarc:
It's choking on bauplan-longbow (and failing to determine the version) because, near as I can tell, there's a bug where `release = false` is only respected *after* determining the version. There's no version to determine for longbow because it's tied to the workspace version.

Setting a nonexistent tag works because it sees that the tag doesn't exist and assumes it's a brand new library; then it sees `publish = false` and skips the package altogether.

I couldn't find a better way to do this. I also updated the changelog configuration to match the docs:

https://release-plz.dev/docs/extra/single-changelog

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -1,6 +1,14 @@
 [workspace]
 git_only = true
 publish = false
+
+[[package]]
+name = "bauplan"
+release = true
+changelog_update = true
+changelog_path = "./CHANGELOG.md"
+git_tag_enable = true
+git_release_enable = true
 git_tag_name = "v{{ version }}"
 git_release_name = "v{{ version }}"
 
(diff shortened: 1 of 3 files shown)


colinmarc:
Example failure: https://github.com/BauplanLabs/bauplan/actions/runs/24802448677/job/72588297982

release-plz.toml
```diff
@@ -1,13 +1,24 @@
...
+changelog_update = true
+changelog_path = "./CHANGELOG.md"
 git_tag_name = "v{{ version }}"
 git_release_name = "v{{ version }}"
 
 [[package]]
 name = "bauplan-longbow"
 release = false
 changelog_update = false
+# This seems to be the only way to get `release-pz release-pr` to ignore
```
bigluck: release-plz

bigluck (approved):

PR #220, opened 2026-04-23
chore: remove openssl as a dev dependency

colinmarc:
I saw dependabot trying to update it (#219) and got spooked. We'd rather not have openssl around.


bigluck (approved):
LGTM

PR #221, opened 2026-04-23
build: tweak release-plz PR name

colinmarc:
This is related to the addition of the bauplan-longbow package. See #218. The current naming would fail gitlint.


bigluck (approved):
<3

PR #223, opened 2026-04-23
fix: statically link the msvc crt in windows binaries

bigluck:
The pysdk already worked on vanilla Windows because python-build-standalone ships `VCRUNTIME140.dll` next to `python.exe`, letting the `.pyd` inherit it at load time.
The standalone `bauplan.exe` had no such host and failed without the `VC++ Redistributable` installed; linking the `MSVC CRT` statically makes the `.exe` self-contained on all native Windows targets.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+# Statically link the MSVC runtime (VCRUNTIME140, MSVCP140, ...) into
+# Windows binaries and cdylibs so the wheel and the standalone .exe
+# work on machines without the Visual C++ Redistributable installed.
+[target.'cfg(all(windows, target_env = "msvc"))']
+rustflags = ["-C", "target-feature=+crt-static"]


colinmarc (approved):
I'm kinda leery of anything that isn't the default, but I have zero idea whether this is correct or not so I'll leave it up to you.

bigluck:
@colinmarc I see, I looked for how `uv` fixed the problem, because when I ran `uv add uv`, it installed `uv.exe` under `Scripts`, and I could run that file (ignore the fact that I also had uv installed on Windows; I'm interested in the binary that's saved in .venv/Scripts.
And searching around, I found this:
- https://github.com/astral-sh/ruff/issues/11503

that's exactly the same error I was getting from cmd, but not from Powershell, which however produced no output

Screenshot 2026-04-23 at 12 51 09

https://github.com/astral-sh/uv/blob/main/.cargo/config.toml#L8

giacomopiccinini (approved):

PR #231, opened 2026-04-27
feat(longbow): stream stdout/stderr from runtimes

colinmarc:
This expands on #197 to add support for attaching to the stdio of running tasks.

You can test with e.g.

```
cargo run -- run -P dev --dry-run --cache off --arg runner2=true
```

### DIFF
--- a/bauplan-longbow/src/client.rs
+++ b/bauplan-longbow/src/client.rs
@@ -10,21 +10,11 @@ use tracing::debug;
 use crate::{ALPN, Error, StreamToken};
 
 struct ArrowStreamState {
-    conn: Connection,
     recv: RecvStream,
     decoder: StreamDecoder,
     remaining: Buffer,
     batch: Option<RecordBatch>,
     _endpoint: Endpoint,
-}
-
-impl Drop for ArrowStreamState {
-    fn drop(&mut self) {
-        // Iroh says to use Endpoint::close, which waits for the
-        // CONNECTION_CLOSE to reach the server. But we don't care, because
-        // we're the only ones receiving data, and the extra latency is stupid.
-        self.conn.close(0_u8.into(), b"done");
-    }
 }
 
 /// An attached user code task. Dropping closes the connection.
@@ -157,8 +147,9 @@ pub async fn fetch_query_results(
             Some(chunk) => {
                 remaining = Buffer::from(chunk.bytes);
                 if let Some(batch) = decoder.decode(&mut remaining)? {
+                    let schema = batch.schema();
                     first_batch = Some(batch);
-                    break decoder.schema().unwrap();
+                    break schema;
                 }
 
                 if let Some(schema) = decoder.schema() {
@@ -173,7 +164,6 @@ pub async fn fetch_query_results(
     };
 
     let state = ArrowStreamState {
-        conn,
         recv,
         decoder,
         remaining,
--- a/bauplan-longbow/src/client.rs
+++ b/bauplan-longbow/src/client.rs
@@ -2,7 +2,7 @@ use arrow::{array::RecordBatch, buffer::Buffer, datatypes::SchemaRef, ipc::reade
 use futures::stream::{self, Stream};
 use iroh::{
     Endpoint, EndpointAddr,
-    endpoint::{ConnectError, Connection, RecvStream, presets::Preset},
+    endpoint::{ConnectError, Connection, ConnectionState, RecvStream, presets::Preset},
 };
 use n0_error::StdResultExt;
 use tracing::debug;
@@ -27,18 +27,20 @@ impl Drop for ArrowStreamState {
     }
 }
 
-/// Connect, sending the desired stream token with the initial handshake
-/// data to shave latency in some scenarios.
+/// An attached user code task. Dropping closes the connection.
+#[derive(Debug)]
+pub struct AttachedTask {
+    /// Stdout from the task.
+    pub stdout: RecvStream,
+    /// Stderr from the task.
+    pub stderr: RecvStream,
+    _endpoint: Endpoint,
+}
+
+/// Connect to an endpoint with the intention to stream user code stdout/stderr.
 ///
-/// If this optimization was successful, the opened stream is returned along
-/// with the connection. Otherwise, it should be opened manually.
-async fn connect(
-    preset: impl Preset,
-    server: EndpointAddr,
-    initial_token: StreamToken,
-) -> Result<(Connection, Endpoint, Option<RecvStream>), Error> {
-    // TODO: we should maybe disable ipv6 proactively, since our servers don't
-    // support it and it can cause extra startup latency. :(
+/// Returns a stream for stdout and stderr, respectively.
+pub async fn attach_task(preset: impl Preset, server: EndpointAddr) -> Result<AttachedTask, Error> {
     let endpoint = Endpoint::builder(preset).bind().await?;
 
     let connecting = endpoint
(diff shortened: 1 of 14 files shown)


bauplan-longbow/src/client.rs
```diff
@@ -27,56 +27,69 @@ impl Drop for ArrowStreamState {
...
+                        stdout: stdout_recv,
+                        stderr: stderr_recv,
+                        _endpoint: endpoint,
+                    });
+                }
                 iroh::endpoint::ZeroRttStatus::Rejected(conn) => {
-                    // The stream doesn't exist from the perspective of
+                    // The streams don't exist from the perspective of
                     // the server.
                     debug!("0rtt rejected");
```
colinmarc, line 126: This code is now duplicated but there was really no way around it.

bauplan-longbow/src/client.rs
```diff
@@ -27,56 +27,69 @@ impl Drop for ArrowStreamState {
     }
 }
 
-/// Connect, sending the desired stream token with the initial handshake
-/// data to shave latency in some scenarios.
+/// An attached user code task. Dropping closes the connection.
```
vtk9, line 27: This part could be a bit more explicit. I initially expected a `impl Drop` somewhere based on this description.

> Dropping closes the connection by closing the 2 streams. `_endpoint` is kept around 

Also, we have a impl Drop for ArrowStream state. does `AttachedTask` need one for the same reason?

bauplan-longbow/src/client.rs
```diff
@@ -27,56 +27,69 @@ impl Drop for ArrowStreamState {
...
+/// Returns a stream for stdout and stderr, respectively.
+pub async fn attach_task(preset: impl Preset, server: EndpointAddr) -> Result<AttachedTask, Error> {
     let endpoint = Endpoint::builder(preset).bind().await?;
 
     let connecting = endpoint
         .connect_with_opts(server, ALPN, Default::default())
         .await
         .map_err(ConnectError::from)?;
 
-    // TODO: since the client is always ephemeral, we have no way to cache
```
vtk9, line 44: I think this TODO was useful - it's still relevant no?

vtk9:
I think we are missing unit tests for this. It should be possible to create a test harness that allows for testing the new code locally without the need for an actual relay server.

vtk9 (approved):

bauplan-longbow/src/client.rs
colinmarc: The endpoint also gets dropped, which closes the connection. I think this is clear enough?

vtk9: I typo'd my initial comment. 

> Dropping closes the connection by closing the 2 streams. _endpoint is kept around *[and dropped last]*

(is what I meant to write) although i should have worded it better

It was not obvious to me until looking through the code and the underlying fields: i.e also that the 2 streams must be dropped before the endpoint gets dropped and so **order** of the fields in `AttachedTask` is important.

colinmarc: Hm, why do you say the ordering is important? I don't think it is, I will double check though.

colinmarc: I see why you're confused, I removed the manual drop on ArrowStreamState in a second commit (I thought I already had).

colinmarc:
Added a test now.

bauplan-longbow/src/client.rs
colinmarc: I went pretty deep on this topic to make sure I wasn't misunderstanding anything.

First of all: `Endpoint::close` is the graceful shutdown of the connection; it closes and waits for the peer to know that it's closed. Simply dropping the `endpoint` doesn't do any graceful shutdown (there's no async drop, of course). In QUIC, graceful shutdown involves sending a special frame (CONNECTION_CLOSE) and waiting for it to be ack'd. Iroh actually has a pretty good explainer for this in their docs: https://www.iroh.computer/blog/closing-a-quic-connection

(Aside: if you're trying to build intuition for why this is more complicated, I think this paragraph from the rustdoc helps:

> Note: Someone used to closing TCP sockets might wonder why it is necessary to wait for timeouts when closing QUIC endpoints, while they don’t have to do this for TCP sockets. This is due to QUIC and its acknowledgments being implemented in user-land, while TCP sockets usually get closed and drained by the operating system in the kernel during the “Time-Wait” period of the TCP socket.

)

In the happy path, we don't really need to perform graceful shutdown. That's because both the client and the server know not to wait for it; the connection is single use, and there are no more clients. So it's perfectly fine to just throw everything away in the name of latency (graceful shutdown involves a rtt).

If the client exits early, the server can be left holding the bag. This can happen if we partially read the query results, for example. The server waits on

```rust
self.send.stopped().await.map_err(|_| Error::StreamClosed)?;
```

Because it never gets any acks from the client, it hits the idle timeout instead (by default 30s). That means in the case that we fetch partial results, we leave the `longbow` runtime running on the server for 30s.

Note that just calling `conn.close()` (which is what the manual `impl Drop` was doing) doesn't actually solve this problem, because nothing is making sure that the close frame actually gets sent/ack'd unless we also call `endpoint.wait_idle()` or `endpoint.close()` (both of which are async, so we can't called it from `Drop`). Sometimes it would work anyway, but it was a race.

However, it's not trivial to implement graceful shutdown because there's no async drop. When we drop the `impl Stream` for arrow batches from the cli/pysdk, we would need to thread the quic connection object all the way up there in order to try and close it, and that still wouldn't fire if the user hit ctrl-c or otherwise killed the process.

Then there's the latency. Connection shutdown in QUIC has this thing where it waits a mandatory amount of time (based on 3x the estimated RTT + variance) for any straggler packets. This is even worse in multipath; the draft RFC that iroh uses specifies you're supposed to use the [max estimated RTT](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-21.html#name-connection-closure) of all the paths in that calculation. Estimation of RTT is a smoothed average and for very short connection, and in our case it can have a high estimate and high variance because it hasn't smoothed yet. So in the worst case this adds 1-3s latency just for teardown, after all the query results are recieved.

All of this is aggravated by the fact that we have very, very short-lived connections.

So, in the end, we have two options:
 - Do graceful shutdown the "proper" way, add latency in basically all cases
 - Do aggressive teardown like we do today, add server-side wait time in the worst case

If we keep the existing logic, we can at least communicate the row limit to the server so that that isn't a problem in the common case.

colinmarc: iroh is fixing this upstream: 

https://github.com/n0-computer/iroh/issues/4201
https://github.com/n0-computer/iroh/pull/4270

src/proto/bpln_proto/commander/service/v2/runner_events.proto
```diff
@@ -52,12 +52,12 @@ message JobFailure {
...
     ERROR_CODE_UNSPECIFIED = 0;
 
-    // Sent when the failure is due to a user error. It can be used to know when
+    // Sent when the failure is due a user error. It can be used to know when
     // to provide more information to the user.
     ERROR_CODE_RUNTIME_TASK_USER_ERROR = 1;
 
     // This is specific to a runtime task failure. If it is unspecified, it
-    // means something wrong happened on scheduler/executor side and should be
+    // means something wrong happend on scheduler/executor side and should be
```
vtk9, line 60: nit: happened mis-spelled

src/cli/run.rs
```diff
@@ -252,7 +258,13 @@ pub(crate) async fn monitor_job_progress(
 
                 handler(event);
             }
-            RunnerEvent::JobCompletion(ev) => return Ok(grpc::interpret_outcome(ev.outcome)?),
+            RunnerEvent::JobCompletion(ev) => {
+                if let Some(ep) = endpoint.get() {
```
vtk9, line 262: maybe we can add a comment here that if for some reason the stdout/stderr from the runtime is delayed, we may technically not receive some logs. up to you

src/proto/bpln_proto/commander/service/v2/runner_events.proto
```diff
@@ -52,12 +52,12 @@ message JobFailure {
     // orchestration of a job
     ERROR_CODE_UNSPECIFIED = 0;
 
-    // Sent when the failure is due to a user error. It can be used to know when
+    // Sent when the failure is due a user error. It can be used to know when
```
vtk9, line 55: i think the due **to** was removed by accident

src/grpc/job.rs
```diff
@@ -267,6 +288,166 @@ impl From<project::ParameterValue> for commanderpb::parameter::Value {
...
-fn pb_to_chrono(ts: prost_types::Timestamp) -> Option<DateTime<Utc>> {
-    Utc.timestamp_opt(ts.seconds, ts.nanos as u32).single()
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum Stdio {
+    Stdout,
+    Stderr,
+}
+
+pub(super) struct JobEventStream {
+    inner: tonic::Streaming<SubscribeLogsResponse>,
```
vtk9, line 306: maybe inner here could be named grpc_events?

src/grpc/job.rs
```diff
@@ -267,6 +288,166 @@ impl From<project::ParameterValue> for commanderpb::parameter::Value {
...
 
-fn pb_to_chrono(ts: prost_types::Timestamp) -> Option<DateTime<Utc>> {
-    Utc.timestamp_opt(ts.seconds, ts.nanos as u32).single()
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum Stdio {
+    Stdout,
+    Stderr,
+}
+
+pub(super) struct JobEventStream {
```
vtk9, line 305: high level nit:

maybe some of change regarding `JobEventStream` these could use extra comments about "why" certain things are done what they are (and/or non obvious gotchas) (some ideas)


```
/// Wraps the gRPC runner-event stream and folds in user-code stdout/stderr,
/// which arrives out-of-band over longbow rather than over gRPC.
///
/// On each `TaskStart` advertising a longbow key, we spawn a reader task that
/// connects to that task's stdout/stderr and forwards lines through an internal
/// channel.
/// `poll_next` then synthesizes `RuntimeUserLog` events from them so
/// downstream consumers see one uniform stream and never learn longbow exists.
pub(super) struct JobEventStream {
```

bauplan-longbow/src/client.rs
```diff
@@ -24,6 +24,23 @@ impl Drop for ArrowStreamState {
     }
 }
 
+/// An attached user code task. Dropping closes the connection.
+#[derive(Debug)]
+pub struct AttachedTask {
```
vtk9, line 29: maybe AttachedTaskHandle. thinking outloud)

bauplan-longbow/src/client.rs
```diff
@@ -74,6 +91,59 @@ async fn connect(
     Ok((conn, recv))
 }
 
+/// Connect to an endpoint with the intention to stream user code stdout/stderr.
+///
+/// Returns a stream for stdout and stderr, respectively.
```
vtk9: nit: maybe 
/// Returns a "handle" which contains the stream for stdout/stderr? Seems strange that we are returning AttachedTask

vtk9 (approved):

src/proto/bpln_proto/commander/service/v2/runner_events.proto
colinmarc: these typos are both copied from the internal protobufs (some llm had fixed them in between :)). I'm inclined to keep them until they're fixed upstream.

src/cli/run.rs
colinmarc: what do you mean delayed?

bauplan-longbow/src/client.rs
colinmarc: they're both pub on the struct, no? not sure what's unclear here.

colinmarc:
I kept all the naming (which is, I think, consistent with the rest of the codebase) but I expanded and updated the comments you pointed out.

PR #235, opened 2026-04-27
fix(longbow): respect --max-rows when fetching query results

colinmarc:
When we introduced longbow query fetching in #197, we didn't respect the max_rows parameter on the CLI or SDK. This fixes that, and also reimplements the row limiting stream logic to be a bit simpler.

### DIFF
--- a/src/flight.rs
+++ b/src/flight.rs
@@ -1,12 +1,6 @@
 //! Support for fetching query results via Arrow Flight.
 
-use std::{
-    sync::{
-        Arc,
-        atomic::{AtomicI64, Ordering},
-    },
-    time,
-};
+use std::{pin::Pin, time};
 
 use arrow::{array::RecordBatch, datatypes::Schema};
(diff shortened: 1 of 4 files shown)


src/flight.rs
```diff
@@ -138,6 +115,30 @@ async fn fetch_batches(
...
 }
 
+/// Truncates a stream of record batches to at most `row_limit` rows total.
+pub fn limit_rows<E>(
+    stream: impl Stream<Item = Result<RecordBatch, E>>,
+    row_limit: Option<u64>,
+) -> impl Stream<Item = Result<RecordBatch, E>> {
+    let remaining = row_limit.unwrap_or(u64::MAX);
+    stream::try_unfold(
+        (Pin::from(Box::new(stream)), remaining),
```
bigluck: nit: `Box::pin(stream)`?

bigluck (approved):

PR #236, opened 2026-04-27
fix: trust OS root certs for ureq HTTPS on debug macOS builds

vtk9:
ureq was using the bundled webpki/Mozilla root bundle, which means
certs trusted by the OS (e.g. CAs in the macOS Keychain) were not
honored. Gate RootCerts::PlatformVerifier behind
cfg(all(debug_assertions, target_os = "macos")) so dev builds on
macOS can talk to services fronted by Keychain-trusted CAs, while
release builds and other platforms keep the Mozilla bundle.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,9 @@ url.workspace = true
 uuid = { version = "1", features = ["serde", "v4"] }
 zip = { version = "4", default-features = false, features = ["deflate"] }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+ureq = { version = "3", features = ["platform-verifier"] }
+
 [build-dependencies]
 anyhow = "1"
 tonic-prost-build = "0.14"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -208,6 +208,16 @@ pub(crate) fn run(args: Args, multiprogress: indicatif::MultiProgress) -> anyhow
 
     // Allows error responses to be parsed.
     let mut cfg = ureq::config::Config::builder().http_status_as_error(false);
+    // Trust macOS Keychain CAs in dev builds so devs can use Keychain-managed
+    // certs (e.g. mkcert) without extra setup.
+    #[cfg(all(debug_assertions, target_os = "macos"))]
+    {
+        cfg = cfg.tls_config(
+            ureq::tls::TlsConfig::builder()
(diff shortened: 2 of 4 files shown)


src/cli.rs
```diff
@@ -207,7 +207,13 @@ pub(crate) fn run(args: Args, multiprogress: indicatif::MultiProgress) -> anyhow
     profile.validate()?;
 
     // Allows error responses to be parsed.
-    let mut cfg = ureq::config::Config::builder().http_status_as_error(false);
+    let mut cfg = ureq::config::Config::builder()
+        .http_status_as_error(false)
+        .tls_config(
```
colinmarc: Let's hide this behind `#[cfg(debug_assertions, target_os = "macos")]`

Cargo.toml
```diff
@@ -107,7 +107,7 @@ tonic-prost = "0.14"
 tracing = { workspace = true, features = ["log"] }
 tracing-indicatif = "0.3"
 tracing-subscriber = { version = "0.3", features = ["chrono", "env-filter"] }
-ureq = "3"
+ureq = { version = "3", features = ["platform-verifier"] }
```
colinmarc: Gate the feature on `degug_assertions, target_os = "macos"`

src/python.rs
```diff
@@ -202,6 +202,16 @@ impl Client {
             .map(time::Duration::from_secs)
             .unwrap_or(time::Duration::from_secs(30));
 
+        #[cfg(all(debug_assertions, target_os = "macos"))]
```
colinmarc, line 208: Can you add a comment about how this is necessary for `mkcert`-based local development?

colinmarc (approved):

src/python.rs
```diff
@@ -202,6 +202,16 @@ impl Client {
...
+        let cfg = ureq::config::Config::builder()
+            .http_status_as_error(false)
+            .tls_config(
+                ureq::tls::TlsConfig::builder()
+                    .root_certs(ureq::tls::RootCerts::PlatformVerifier)
+                    .build(),
+            )
+            .timeout_global(Some(client_timeout));
+        #[cfg(not(all(debug_assertions, target_os = "macos")))]
```
colinmarc: I also think you can just do `mut config` rather than having the `cfg(not)` branch. or `if cfg!(..) {`.

colinmarc (approved):

PR #239, opened 2026-05-05
feat: Type Contracts

drin:
This is a PR to add support for type contracts.

The intention is for new syntax to be gated behind a feature flag (`__bpln_feature_typecontracts__`). The new syntax does not "do" much, meaning that at runtime the new syntax should be functional:

* `Parameter` will return `None` if there's no value associated with the requested parameter
* `ModelTask` (functions decorated with `@bauplan.model`) can still be invoked in the usual way and have all of the usual metadata set (the function wrapper inherits the decorated function's metadata, etc.)

The caveat with `ModelTask` is that because it assigns the decorated function to the `__call__` method of the `ModelTask` proxy object, it won't inspect the same way as a pure function (but you can inspect the `__call__` method and it should be the same).

The goal is to add some logic and structures into the SDK that can be leveraged at runtime or at least adds support for new syntax that can be parsed. This should not affect how the SDK is used today and it should not have significant impact on existing code even if the new syntax is adopted (everything should be transparent).

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -143,7 +143,15 @@ pub(crate) fn job_request_common(
     priority: Option<Priority>,
 ) -> commanderpb::JobRequestCommon {
     let hostname = gethostname::gethostname().to_string_lossy().into_owned();
-    let args = args.into_iter().map(KeyValue::into_strings).collect();
+    let mut args: std::collections::HashMap<String, String> =
+        args.into_iter().map(KeyValue::into_strings).collect();
+
+    if std::env::var("BPLN_ENABLE_TYPE_CONTRACT")
+        .map(|v| v.eq_ignore_ascii_case("true") || v == "1")
+        .unwrap_or(false)
+    {
+        args.insert("enable_type_contracts".to_owned(), "true".to_owned());
+    }
 
     commanderpb::JobRequestCommon {
         module_version: env!("CARGO_PKG_VERSION").to_owned(),
--- a/tests/cli/run.rs
+++ b/tests/cli/run.rs
@@ -259,6 +259,55 @@ fn multiparent() {
 //         .stdout(contains("s3write parallel_child_1"));
 // }
 
+// This test will fail until the server is able to install an SDK with the new feature flag.
+#[test]
+#[ignore]
+fn typed_parameters_idempotent() {
+    // The fixture monkey-patches __bpln_feature_typecontracts__ to True, so this
+    // test exercises the typed-parameter code path regardless of the compile-time flag.
+    bauplan()
+        .args([
+            "run",
+            "--cache",
+            "off",
+            "--dry-run",
+            "-p",
+            "tests/fixtures/typed-parameters",
+            "--param",
+            "location_id=123",
+            "--param",
+            "golden_ratio=4.2",
+            "--param",
+            "use_random_forest=false",
+            "--param",
+            "start_datetime=2023-01-01T00:00:00+00:00",
+            "--param",
+            "end_datetime=2023-01-02T00:00:00+00:00",
+        ])
+        .assert()
+        .success()
+        .stderr(contains("yayparams.num_rows=629154"))
+        .stderr(contains("yayparams.num_columns=3"));
+}
+
+#[test]
+fn typed_parameters_feature_flag() {
+    // The old Parameter('name') syntax works in both modes. When type contracts
+    // are enabled, a deprecation warning is emitted but the run still succeeds.
+    let assertion = bauplan()
+        .args([
+            "run",
+            "--cache",
+            "off",
+            "--dry-run",
+            "-p",
+            "tests/fixtures/parameters",
+        ])
+        .assert();
+
+    assertion.success().stderr(contains("golden_ratio=1.666"));
+}
+
 #[test]
 fn parameters_project() {
     bauplan()
--- a/tests/fixtures/typed-parameters/models.py
+++ b/tests/fixtures/typed-parameters/models.py
@@ -1,0 +1,43 @@
+# Force the type-contracts feature on regardless of the compile-time flag,
+# so that this fixture can be used as an idempotent test.
+# Must be set before bauplan._internal is imported.
+import os
+
+os.environ['BPLN_ENABLE_TYPE_CONTRACT'] = 'True'
+
+import bauplan
+import pyarrow as pa
+
+from bauplan import Parameter
+
+
+@bauplan.model(
+    columns=['y'],
+    materialization_strategy='NONE',
+)
+@bauplan.python('3.13')
+def typed_params_model(
+    yayparams=bauplan.Model(  # noqa: B008, ANN001
+        'query_model',
+        columns=[
+            'dropoff_datetime',
+            'pickup_datetime',
+            'trip_miles',
+        ],
+        filter='PULocationID = $location_id',
+    ),
+    golden_ratio=Parameter.golden_ratio,
+    use_random_forest=Parameter.use_random_forest,
+    start_datetime=Parameter.start_datetime,
+    end_datetime=Parameter.end_datetime,
+) -> pa.Table:
+    print(f'golden_ratio={golden_ratio}')
+    print(f'use_random_forest={use_random_forest}')
+    print(f'start_datetime={start_datetime}')
+    print(f'end_datetime={end_datetime}')
+
+    print(yayparams)
+    print(f'yayparams.num_rows={yayparams.num_rows}')
+    print(f'yayparams.num_columns={yayparams.num_columns}')
+
+    return pa.Table.from_pydict({'y': [1, 2, 3]})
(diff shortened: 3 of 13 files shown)


drin:
I guess I'm not supposed to bump the version?

tests/cli.rs
```diff
@@ -27,8 +27,40 @@ pub fn bauplan() -> assert_cmd::Command {
     cmd
 }
 
-pub fn username() -> String {
-    std::env::var("BPLN_USERNAME").unwrap_or_else(|_| "bauplan-e2e-check".to_string())
+#[cfg(feature = "_interactive")]
```
vtk9: what does this do? is this relevant for interactive vs non-interactive shells? is it needed this PR?

src/cli/run.rs
```diff
@@ -143,7 +143,15 @@ pub(crate) fn job_request_common(
     priority: Option<Priority>,
 ) -> commanderpb::JobRequestCommon {
     let hostname = gethostname::gethostname().to_string_lossy().into_owned();
-    let args = args.into_iter().map(KeyValue::into_strings).collect();
+    let mut args: std::collections::HashMap<String, String> =
+        args.into_iter().map(KeyValue::into_strings).collect();
+
+    if std::env::var("BPLN_ENABLE_TYPE_CONTRACT")
```
vtk9, line 149: this code path is somewhat duplicated (only twice though) so up to you whether or not to change

```
pub fn type_contracts_enabled() -> bool {
    std::env::var("BPLN_ENABLE_TYPE_CONTRACT")
        .map(|v| v.eq_ignore_ascii_case("true") || v == "1")
        .unwrap_or(false)
}
```

tests/cli/run.rs
```diff
@@ -259,6 +259,53 @@ fn multiparent() {
 //         .stdout(contains("s3write parallel_child_1"));
 // }
 
+#[test]
+fn typed_parameters_idempotent() {
+    // The fixture monkey-patches __bpln_feature_typecontracts__ to True, so this
```
vtk9, line 266: > The fixture monkey-patches
Is this still accurate?

tests/cli.rs
drin: I use it to test just so I don't have to set both BPLN_PROFILE and BPLN_USERNAME. Instead, username can be derived from the active profile.

I suppose it isn't needed in this PR but it does make usage a bit easier by reducing the number of env variables needed.

src/cli/run.rs
drin: you're right. I can abstract it

tests/cli/run.rs
drin: yes, it's done on line 6 of `tests/fixtures/typed-parameters/models.py`. Although it's done via setting an environment variable before importing the library (so.. it's not exactly **monkey-patching**)

tests/cli.rs
drin: mostly it makes it easier to switch between envs because I can just switch profile names instead of also having to remember what username is associated with each profile name

src/cli/run.rs
colinmarc: Why have this be an environment variable at all? You already have the arg.

tests/cli.rs
colinmarc: I'm in favor of dropping BPLN_USERNAME altogether, I think this is leftover from the old e2e test harness. Not in favor of adding a feature flag.

colinmarc: (but do it in a separate PR)

tests/fixtures/typed-parameters/models.py
```diff
@@ -0,0 +1,43 @@
...
+    yayparams=bauplan.Model(  # noqa: B008, ANN001
+        'query_model',
+        columns=[
+            'dropoff_datetime',
+            'pickup_datetime',
+            'trip_miles',
+        ],
+        filter='PULocationID = $location_id',
+    ),
+    golden_ratio=Parameter.golden_ratio,  # noqa: B008, ANN001
```
colinmarc: It's a nonstarter to disable typechecking on our fancy new type contracts

pyproject.toml
```diff
@@ -95,6 +95,9 @@ exclude = [
     "examples",
 ]
 
+[tool.ruff.format]
+quote-style = "preserve"
```
colinmarc: ?

tests/fixtures/typed-parameters/models.py
colinmarc: This also says "the default value for golden_ratio is `Parameter.golden_ratio`" which makes no sense to me. I thought we were typing parameters, not using default values anymore?

pyproject.toml
drin: I just didn't want to switch everything to double quote (I prefer single quote in python for strings). I figured this lint change would be relatively non-controversial

src/cli/run.rs
drin: I think the run arg doesn't get captured in the SDK unless a run command is issued? As an environment variable I can enable some features at runtime (activated by runner/runtime) in the SDK

pyproject.toml
colinmarc: Nah, let's not change the ruff rules in this PR. And I'm strongly inclined to keep the defaults since it avoids the possibility of config mismatch.

src/cli/run.rs
colinmarc: args are passed to every method - what do you need to do outside a method?

drin: but the runtime doesn't use the SDK through the run command. when the runtime imports bauplan and imports a user's code snapshot, are run arguments available somehow?

colinmarc: Arguments are available to every method call. I'm not seeing how this env variable is even available on the runtime though.

drin: in another PR that is dependent on this one (PR#5098)

tests/fixtures/typed-parameters/models.py
drin: Ultimately, I think the assignment op has a connotation that's useful and we want to support syntax that is much easier with default values, so I don't think we're trying to remove them entirely.

Separately, in the case of parameters, there's simply no way to enable typechecking because it's they're defined in configuration instead of code. Unless it is also defined in code, the best case scenario is for some parsing logic to set the type. The `TypeRegistryMeta` metaclass would allow this functionality (I experimented with it but removed it because `Parameter['golden_ratio']` as a type was confusing to some people).

For now, this proposed syntax (`Parameter.golden_ratio`) ends up being a slightly less cluttered mechanism than the existing syntax (`Parameter('golden_ratio')`), but provides a mildly easier implementation path to a well-typed representation (because it calls a function, `__class_getattribute__`, which can have a Generic return type instead of customizing the __new__ implementation).

colinmarc: Parameters are stored locally, so they're accessible to the typechecker in theory. But my point is more that anything we do with the new types has to pass muster with `ty`, or it's not worth doing imo.

tests/cli.rs
drin: okay, moved to separate PR

tests/fixtures/typed-parameters/models.py
drin: but between gradual typing and the difference between typing parameters and tables I don't think this is a non starter?

drin: I guess you're just saying that you're not convinced the change to parameters syntax is worth it because it's not a clear improvement on type annotation (despite my reasoning)?

src/cli/run.rs
drin: adding the function to `src/config.rs` for checking if the feature is enabled.

tests/fixtures/typed-parameters/models.py
drin: I removed the noqa comments. I didn't realize that's all you were talking about

vtk9:
LGTM. Approving as:
- we don't have an easy way to test this internally without many workarounds. we are blocked on internal testing until this is merged and released
- it's behind a feature flag and not enabled by default and can still be changed

CICD is failing but @drin has a step forward

vtk9 (approved):

drin:
CICD passing now. There is a test that I added, `typed_parameters_idempotent`, that uses the test fixture `tests/fixtures/typed-parameters. This test is "problematic" because it always runs with the feature turned on (hence, idempotent); but, until this PR is merged (and released) no server is able to install an SDK that can turn the feature on.

So, in this PR I've marked the test as ignore so that it can be enabled (if desired) in a future PR; otherwise, it can be enabled locally.

I've also removed comments for pyright because `ty` doesn't care about it and just logs warnings

colinmarc:
> I've also removed comments for pyright because `ty` doesn't care about it and just logs warnings

That's only because `tests/fixtures` is explicitly ignored in `pyproject.toml`. The file does not pass typechecking.

Reverted in https://github.com/BauplanLabs/bauplan/pull/265

PR #279, opened 2026-05-29
feat: add nested SQL

giacomopiccinini:
This PR allows users to extend a `bauplan_project.yml` with an additional parameter `include_paths`. 
`include_paths` is a list of (relative) glob patterns which will trigger the inclusion of SQL files _only_ in the code snapshot. 

Current limitations:
- Only SQL files can be included (no Python, or others)
- Research can only happen "downward" from the Bauplan project directory; upward search is prohibited.
- Gitignored files are ignored, regardless of them being included in `include_paths`. 

Loud failure modes:
1. Pattern does not end with `.sql` (_not_ case sensitive);
2. Pattern is absolute and not relative;
3. Pattern points outside project directory;
4. Patterns including `..` are not allowed;
5. Non UTF-8 patterns;
7. File extension is not `.sql` (case sensitive);
8. File is a symlink pointing outside the project directory. 

Tested on Windows & MacOS

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,3 +33,16 @@ jobs:
         env:
           BAUPLAN_API_ENDPOINT: https://api.use1.aqa.bauplanlabs.com
           BAUPLAN_API_KEY: ${{ secrets.BAUPLAN_API_KEY }}
+
+  # Test path handling on windows, but don't bother with the entire integration test suite.
+  test-windows:
+    name: Unit Tests (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+      - uses: ./.github/actions/setup-caches
+
+      - run: mise x -- cargo test --lib
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,8 @@ ureq = { version = "3", features = ["platform-verifier"] }
 url.workspace = true
 uuid = { version = "1", features = ["serde", "v4"] }
 zip = { version = "4", default-features = false, features = ["deflate"] }
+globset = "0.4.18"
+ignore = "0.4.26"
 tokio-util = { version = "0.7.18", features = ["codec"] }
 
 [build-dependencies]
--- a/src/project.rs
+++ b/src/project.rs
@@ -1,8 +1,8 @@
 //! Helpers for managing bauplan projects.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::io::{Read, Write};
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
 use base64::Engine;
 use rsa::sha2::Sha256;
@@ -10,6 +10,9 @@ use rsa::{Oaep, RsaPublicKey};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use uuid::Uuid;
+
+use globset::{Glob, GlobSetBuilder};
+use ignore::WalkBuilder;
 
 /// Errors that can occur when working with project files.
 #[derive(Debug, Error)]
@@ -27,6 +30,16 @@ pub enum ProjectError {
     Zip(#[from] zip::result::ZipError),
     #[error("encryption failed")]
     Encryption(#[from] rsa::Error),
+    #[error("invalid glob pattern")]
+    Glob(#[from] globset::Error),
+    #[error("failed to resolve include paths")]
+    Ignore(#[from] ignore::Error),
+    #[error("glob pattern not allowed: {0}")]
+    GlobPatternNotAllowed(String),
+    #[error("symlink not allowed: {0} points to {1}")]
+    Symlink(PathBuf, PathBuf),
+    #[error("failed to extract relative path")]
+    Prefix(#[from] std::path::StripPrefixError),
     #[error("invalid value {0:?} of type {1}")]
     InvalidParameterValue(String, ParameterType),
 }
@@ -256,6 +269,12 @@ pub struct ProjectFile {
     /// Parameters for models.
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub parameters: BTreeMap<String, ParameterDefault>,
+    // /// Additional paths to include as glob expressions.
+    // #[serde(default, skip_serializing_if = "Option::is_none")]
+    // pub include_paths: Option<Vec<String>>,
+    /// Additional paths to include as glob expressions.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub include_paths: Vec<String>,
 
     /// The location of the project file on disk.
     #[serde(skip)]
@@ -302,27 +321,33 @@ impl ProjectFile {
             )
         })?;
 
+        // Additional, user-provided patterns.
+        let additional_patterns = self
+            .include_paths
+            .iter()
+            .map(|p| resolve_pattern(p))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let files: HashSet<PathBuf> =
+            resolve_includes(project_dir, &additional_patterns)?.collect();
+
         let mut buf = Vec::new();
         let mut zip = zip::ZipWriter::new(std::io::Cursor::new(&mut buf));
         let options = zip::write::SimpleFileOptions::default()
             .compression_method(zip::CompressionMethod::Deflated);
 
         let mut contents = Vec::new();
-        for entry in std::fs::read_dir(project_dir)? {
-            let path = entry?.path();
-            let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
-                continue;
-            };
-
-            if !include_in_snapshot(name) {
-                continue;
-            }
+        for path in files {
+            // The zip spec mandates forward slashes in entry names, and serializing
+            // the Path would produce backslashes on Windows; start_file_from_path
+            // does the conversion.
+            let name = path.strip_prefix(project_dir)?;
 
             contents.clear();
             let mut file = std::fs::File::open(&path)?;
             file.read_to_end(&mut contents)?;
 
-            zip.start_file(name, options)?;
+            zip.start_file_from_path(name, options)?;
             zip.write_all(&contents)?;
         }
 
@@ -331,10 +356,132 @@ impl ProjectFile {
     }
 }
 
-fn include_in_snapshot(name: &str) -> bool {
-    name.ends_with(".py")
-        || name.ends_with(".sql")
-        || name == "requirements.txt"
-        || name == "bauplan_project.yml"
-        || name == "bauplan_project.yaml"
+/// Given a glob pattern, ensure the pattern is "admissible".
+fn resolve_pattern(p: &str) -> Result<String, ProjectError> {
+    // Users should be explicitly including file extensions, not globbing for all files.
+    if !p.ends_with(".sql") {
+        return Err(ProjectError::GlobPatternNotAllowed(p.to_string()));
+    }
+
+    // Absolute patterns are not allowed, only relative.
+    let pattern = Path::new(p);
+    if pattern.is_absolute() {
+        return Err(ProjectError::GlobPatternNotAllowed(format!(
+            "{} is an absolute path",
+            p
+        )));
+    }
+
+    // Patterns with ".." are not allowed.
+    if pattern
+        .components()
+        .any(|c| matches!(c, Component::ParentDir))
+    {
+        return Err(ProjectError::GlobPatternNotAllowed(p.to_string()));
+    }
+
+    // Strip leading ./ and use forward slashes for glob syntax.
+    Ok(p.strip_prefix("./").unwrap_or(p).replace('\\', "/"))
+}
+
+/// Include files in snapshot, based on provided glob patterns.
+/// Ignoring takes precedence over inclusion: any ignored file will not be included
+/// regardless of inclusion patterns.
+fn resolve_includes<S: AsRef<str>>(
+    base: &Path,
+    patterns: &[S],
+) -> Result<impl Iterator<Item = PathBuf>, ProjectError> {
+    let base_canonical = base.canonicalize()?;
+
+    // Build a glob set, matching all the required patterns.
+    let mut glob_builder = GlobSetBuilder::new();
+
+    // Top level patterns, always included.
+    let base_patterns = [
+        "*.py",
+        "*.sql",
+        "requirements.txt",
+        "bauplan_project.yml",
+        "bauplan_project.yaml",
+    ];
+
+    for p in patterns {
+        glob_builder.add(Glob::new(p.as_ref())?);
+    }
+
+    for p in base_patterns {
+        glob_builder.add(Glob::new(p).unwrap());
+    }
+
+    let set = glob_builder.build()?;
+
+    let mut paths = BTreeSet::new();
+
+    for entry in WalkBuilder::new(&base_canonical).build() {
+        let entry = entry?.into_path();
+        let canonical = entry.canonicalize()?;
+
+        if !canonical.is_file() {
+            continue;
+        }
+
+        let rel_entry = canonical
+            .strip_prefix(&base_canonical)
+            .map_err(|_| ProjectError::Symlink(entry, canonical.clone()))?
+            .to_path_buf();
+
+        // For globset to work, we need relative paths.
+        if !set.is_match(&rel_entry) {
+            continue;
+        }
+
+        // For zipping to work, we need absolute paths.
+        paths.insert(canonical);
+    }
+
+    Ok(paths.into_iter())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn resolve_pattern_rejects_upward_pattern() {
+        let tmp = tempfile::tempdir().unwrap();
+        let proj = tmp.path().join("proj");
+        std::fs::create_dir_all(proj.join("views")).unwrap();
+        std::fs::write(proj.join("views").join("age.sql"), "SELECT 1").unwrap();
+
+        let result = resolve_pattern("../proj/views/*.sql");
+        assert!(matches!(
+            result,
+            Err(ProjectError::GlobPatternNotAllowed(..))
+        ));
+    }
+
+    fn symlink_file(src: &Path, dst: &Path) {
+        #[cfg(unix)]
+        std::os::unix::fs::symlink(src, dst).unwrap();
+        #[cfg(windows)]
+        std::os::windows::fs::symlink_file(src, dst).unwrap();
+    }
+
+    #[test]
+    fn resolve_includes_rejects_symlink_escaping_base() {
+        let tmp = tempfile::tempdir().unwrap();
+        let proj = tmp.path().join("proj");
+        std::fs::create_dir_all(proj.join("views")).unwrap();
+
+        std::fs::write(tmp.path().join("secret.sql"), "SELECT secret").unwrap();
+
+        symlink_file(
+            &tmp.path().join("secret.sql"),
+            &proj.join("views").join("leak.sql"),
+        );
+
+        let patterns = &[resolve_pattern("views/*.sql").unwrap()];
+        let result = resolve_includes(&proj, patterns);
+        assert!(matches!(result, Err(ProjectError::Symlink(..))));
+    }
 }
--- a/tests/cli/run.rs
+++ b/tests/cli/run.rs
@@ -490,3 +490,66 @@ fn with_transaction() {
         .success()
         .stderr(contains("num_rows= 430488"));
 }
+
+#[test]
+fn sql_nested() {
+    bauplan()
+        .args([
+            "run",
+            "--dry-run",
+            "--cache",
+            "off",
+            "-p",
+            "tests/fixtures/sql_nested",
+        ])
+        .assert()
+        .success();
+}
+
+#[test]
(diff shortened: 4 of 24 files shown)


src/project.rs
```diff
@@ -338,3 +362,71 @@ fn include_in_snapshot(name: &str) -> bool {
...
 }
+
+fn include_paths_filter(file: &Path) -> bool {
+    file.extension()
+        .and_then(OsStr::to_str)
+        .map(str::to_ascii_lowercase)
+        .is_some_and(|ext| matches!(ext.as_str(), "sql")) // Future: extend with "sql" | "py"
+}
+
+fn resolve_includes<F>(
```
bigluck: I'd like to re-review the resolution rules for `include_paths` here, because right now the code drops anything that does not match without telling the user, and this is the thing I really want to avoid because it could introduce bugs now or in future versions of the cli.

First, I think it's safer to be additive and not a replacement.
The root files (`./*.sql`, `./*.py`, the project file and `requirements.txt`) should always go into the snapshot, and `include_paths` only adds extra folders on top.
In this way, if a user adds one SQL subfolder but forgets to also list `./*.py` / `./bauplan_project.yml`, they do not lose their root files.
And when `include_paths` is not set, the behavior stays the same as today.

Second, I want it explicit and failing loudly, never in silence. The user has to write the extension in the pattern. `.sql` can be at any depth, `.py` only in the root for now.
Everything else should give a clear client-side error that mentions it as future work, instead of being filtered out quietly. So a bare glob without extension like `./**/*` or `./folder/*` is an error, a `.py` outside the root like `./folder/*.py` is an error, and the same for upward paths like `../*.sql` or extensions I do not support like `./*.csv`.

I ask for the extension because of the future. A bare `./**/*` today takes only SQL, but the day we add more file types its meaning changes without the user touching anything.

src/project.rs
```diff
@@ -302,21 +312,35 @@ impl ProjectFile {
...
-            let path = entry?.path();
-            let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
-                continue;
-            };
-
-            if !include_in_snapshot(name) {
-                continue;
-            }
+        for path in files {
+            let name = path
```
bigluck: I'm not 100% sure about it, but on winzoz strip_prefix keeps the `\` separator, so a nested file is zipped as `views\age.sql` and the server  reads it as a flat file.

bigluck (commented):
this is my first pass

src/project.rs
colinmarc: Let's also fail loudly if an included file would be gitignored. The `ignore` crate is a good one here.

tests/cli/run.rs
```diff
@@ -490,3 +490,33 @@ fn with_transaction() {
...
+#[test]
+fn sql_prohibited() {
+    bauplan()
+        .args([
+            "run",
+            "--dry-run",
+            "--cache",
+            "off",
+            "-p",
+            "tests/fixtures/sql_prohibited",
```
bigluck: Be careful, this test passes for the wrong reason. `bauplan run` expects to point to a folder with a `bauplan_project.yaml` file in it, but here you are pointing to the folder where the `pipeline/` folder is, so in reality the request fails because it does not find a project

giacomopiccinini: yeah i know, this was fixed. just haven't pushed the code yet.

src/project.rs
giacomopiccinini: done

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+        .map(str::to_ascii_lowercase)
+        .is_some_and(|ext| matches!(ext.as_str(), "sql")) // Future: extend with "sql" | "py".
+}
+
+/// Include additional files in snapshot, based on user-provided glob patterns.
+fn resolve_includes(base: &Path, patterns: &[String]) -> Result<Vec<PathBuf>, ProjectError> {
+    let base_canonical = base.canonicalize()?;
+
+    let mut builder = GlobSetBuilder::new();
+    for p in patterns {
```
bigluck: Can we fail if an include pattern matches zero files? for example `include_paths: ["view/*.sql"]` currently succeeds silently, but the intended sql model is missing and later shows up as “table not found”

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+        builder.add(Glob::new(&glob)?);
+    }
+
+    let set = builder.build()?;
+
+    let mut paths = HashSet::new();
+    let mut walked_gitignores = Vec::new();
+
+    for entry in WalkDir::new(&base_canonical) {
+        let entry_canonical = entry?.path().canonicalize()?;
```
bigluck: If the project has `include_paths: ["views/*.sql"]` plus an unrelated broken symlink or symlink to, idk, `/tmp/blablablabal.sql` under a different folder `ciaociao/`, I think `canonicalize()` may fail or reject it before we know it does not match the glob

tests/cli/run.rs
```diff
@@ -490,3 +490,82 @@ fn with_transaction() {
...
+            "tests/fixtures/sql_prohibited_extension",
+        ])
+        .assert()
+        .failure()
+        .stderr(contains("glob pattern not allowed"));
+}
+
+#[test]
+fn sql_upward_inside() {
+    // The patterns climb out of base but resolve back inside it, so they are valid
```
bigluck: can you update the pr body/docs to match this new behavior? It says upward search is prohibited, but this test intentionally allows `../sql_upward_inside/...` when it resolves back inside the project

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+}
+
+/// Include additional files in snapshot, based on user-provided glob patterns.
+fn resolve_includes(base: &Path, patterns: &[String]) -> Result<Vec<PathBuf>, ProjectError> {
+    let base_canonical = base.canonicalize()?;
+
+    let mut builder = GlobSetBuilder::new();
+    for p in patterns {
+        // Users should be explicitly including .sql files, not globbing for all files.
+        if !p.ends_with(".sql") {
```
bigluck: should pattern validation be case-insensitive too? `include_paths_filter` accepts `model.SQL`, but at `include_paths: ["views/*.SQL"]` is rejected

bigluck (commented):
2nd pass

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+            return Err(ProjectError::GlobPatternNotAllowed(p.to_string()));
+        }
+
+        // Resolve the pattern lexically against base. Patterns may climb upward
+        // as long as they resolve back inside base, e.g. ../<base dirname>/views/*.sql.
+        // Components are walked manually instead of using join: on Windows the
+        // canonicalized base is \\?\-prefixed and join would resolve `..` itself,
+        // popping wildcards before the guard below can see them.
+        let pattern = Path::new(p);
```
bigluck: should we reject absolute paths? even if `/abs/project/views/*.sql` resolves inside the project (on the user's computer) it makes `bauplan_project.yaml` machine specific; I’d suggest to have in `include_paths` only project-relative paths.

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+
+        // Collect walked .gitgnores for further processing.
+        if entry_canonical.file_name() == Some(OsStr::new(".gitignore")) {
+            walked_gitignores.push(entry_canonical);
+            continue;
+        }
+
+        let relative = entry_canonical.strip_prefix(&base_canonical)?;
+        if set.is_match(relative) {
+            if !include_paths_filter(&entry_canonical) {
```
bigluck: I'm not 100% sure but isn't this branch dead code?

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+
+        builder.add(Glob::new(&glob)?);
+    }
+
+    let set = builder.build()?;
+
+    let mut paths = HashSet::new();
+    let mut walked_gitignores = Vec::new();
+
+    for entry in WalkDir::new(&base_canonical) {
```
bigluck: i'm a bit wary that this walks the entire tree (including `.git`, `.venv`...) and canonicalizes every file just to match a couple of globs, but maybe it's unavoidable since we also need to collect the nested `.gitignore` files in the same pass?

src/project.rs
```diff
@@ -302,27 +327,40 @@ impl ProjectFile {
             )
         })?;
 
+        let mut files: HashSet<PathBuf> = HashSet::new();
+
+        // Top level first, always included.
+        for entry in std::fs::read_dir(project_dir)? {
```
bigluck: one thing I'm not sure about: the `.gitignore` check runs inside `resolve_includes`, so it covers the `include_paths` files but not the top-level ones we always add here.
If someone keeps a `local_secrets.py` or a generated `.py` in the project root and `.gitignores` it, wouldn't it still end up in the snapshot silently, while a `.gitignored` `views/*.sql` would instead hard-fail?

Cargo.toml
```diff
@@ -111,6 +111,10 @@ ureq = { version = "3", features = ["platform-verifier"] }
 url.workspace = true
 uuid = { version = "1", features = ["serde", "v4"] }
 zip = { version = "4", default-features = false, features = ["deflate"] }
+glob = "0.3.3"
```
bigluck: do we still need `glob` here?

src/project.rs
```diff
@@ -338,3 +376,321 @@ fn include_in_snapshot(name: &str) -> bool {
...
+
+        for (name, expected) in cases {
+            assert_eq!(
+                include_paths_filter(Path::new(name)),
+                expected,
+                "wrong classification for extension of {name}"
+            );
+        }
+    }
+}
```
bigluck: rustfmt?

src/project.rs
```diff
@@ -302,27 +327,40 @@ impl ProjectFile {
...
-            };
-
-            if !include_in_snapshot(name) {
-                continue;
-            }
+        for path in files {
+            // The zip spec mandates forward slashes in entry names, and serializing
+            // the Path would produce backslashes on Windows; start_file_from_path
+            // does the conversion
+            let name = path.strip_prefix(project_dir).unwrap_or(&path);
```
bigluck: I’d avoid the silent fallback here.. if `strip_prefix(project_dir)` fails, shouldn’t we return an error instead of potentially writing an absolute or non-project-relative entry into the zip?

Cargo.toml
giacomopiccinini: Nope, removed.

src/project.rs
giacomopiccinini: done

src/project.rs
giacomopiccinini: That's correct. I guess the point is that it would alter the current behavior which I don't think we want?

src/project.rs
giacomopiccinini: Good point, fixed.

src/project.rs
giacomopiccinini: testing this.

src/project.rs
giacomopiccinini: fixed I think, to be tested on Windows

src/project.rs
giacomopiccinini: i think this is the price for having explicit failures for gitignored files (plus a bunch of other loud failures). one could use e.g. the `ignore` crate walker: that would filter out automatically the ignored files but then we would not be able to fail loudly (i think)

giacomopiccinini: picking up the `.gitignore` is yet another problem

src/project.rs
giacomopiccinini: I don't think it is? based on what we said before

src/project.rs
giacomopiccinini: I guess we could do that if we treated the patterns separately. But with `builder.add(Glob::new(&glob)?);` we are actually adding all the globs, then buildingin, and then looping through files. With the curent setup i don't think we can really do that

tests/cli/run.rs
giacomopiccinini: done

src/project.rs
colinmarc: I think we need to use the `ignore` crate walker; this will also take ten minutes on rust projects with a 60gb target directory.

.github/workflows/tests.yaml
```diff
@@ -33,3 +33,18 @@ jobs:
...
+  # code (path handling, globs, zip entry names) that an ubuntu runner can't.
+  test-windows:
+    name: Unit Tests (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+
```
colinmarc: nit: remove these extra newlines

.github/workflows/tests.yaml
```diff
@@ -33,3 +33,18 @@ jobs:
         env:
           BAUPLAN_API_ENDPOINT: https://api.use1.aqa.bauplanlabs.com
           BAUPLAN_API_KEY: ${{ secrets.BAUPLAN_API_KEY }}
+
+  # Unit tests only: no backend access, but they cover the platform-sensitive
```
colinmarc: This comment is kinda confusing. I think something like this would be better:

```
# Test path handling on windows, but don't bother with the entire integration test suite.
```

src/project.rs
colinmarc: I would actually go for changing the behavior. We can configure the [max depth](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.max_depth) on the scan and use the same basic logic whether or not there are extra paths.

src/project.rs
```diff
@@ -338,3 +375,209 @@ fn include_in_snapshot(name: &str) -> bool {
...
+        glob.push_str(part);
+    }
+
+    Ok(glob)
+}
+
+/// Include additional files in snapshot, based on user-provided glob patterns.
+/// Ignoring takes precedence over inclusion: any ignored file will not be included
+/// regardless of `include_paths` patterns.
+fn resolve_includes(base: &Path, patterns: &[String]) -> Result<Vec<PathBuf>, ProjectError> {
```
colinmarc: nit:


```suggestion
fn resolve_includes(base: &Path, patterns: &[String]) -> Result<impl Iterator<Item=PathBuf>, ProjectError> {
```

And drop the collect at the end.

src/project.rs
```diff
@@ -338,3 +375,209 @@ fn include_in_snapshot(name: &str) -> bool {
...
+    // Build a glob set, matching all the included paths.
+    let mut glob_builder = GlobSetBuilder::new();
+
+    for p in patterns {
+        let resolved_pattern = resolve_pattern(&base_canonical, p)?;
+        glob_builder.add(Glob::new(&resolved_pattern)?);
+    }
+    let set = glob_builder.build()?;
+
+    let mut paths = HashSet::new();
```
colinmarc:
```suggestion
    let mut paths = BtreeSet::new();
```

src/project.rs
```diff
@@ -338,3 +375,209 @@ fn include_in_snapshot(name: &str) -> bool {
...
+/// and resolve relative components (.., etc.).
+fn resolve_pattern(base_canonical: &Path, p: &String) -> Result<String, ProjectError> {
+    // Users should be explicitly including file extensions, not globbing for all files.
+    // Also, we do not enforce constraints on lower/upper case here.
+    if !p.to_lowercase().ends_with(".sql") {
+        return Err(ProjectError::GlobPatternNotAllowed(p.to_string()));
+    }
+
+    // Resolve the pattern lexically against base. Patterns may climb upward
+    // as long as they resolve back inside base, e.g. ../<base dirname>/views/*.sql.
```
colinmarc: Is there a really good reason to support that? In my opinion globs with `..` are kinda weird and I would be surprised if lots of tooling supports that.

src/project.rs
```diff
@@ -338,3 +375,209 @@ fn include_in_snapshot(name: &str) -> bool {
...
+        let entry = entry?.into_path();
+        let canonical = entry.canonicalize()?;
+        if !canonical.is_file() {
+            continue;
+        }
+        let rel_entry = canonical
+            .strip_prefix(&base_canonical)
+            .map_err(|_| ProjectError::Symlink(entry, canonical.clone()))?
+            .to_path_buf();
+        // For globset to work, we need relative paths.
```
colinmarc, line 433: nit: newline before each of the `if` checks.

src/project.rs
```diff
@@ -338,3 +375,209 @@ fn include_in_snapshot(name: &str) -> bool {
...
+    for entry in WalkBuilder::new(&base_canonical).build() {
+        let entry = entry?.into_path();
+        let canonical = entry.canonicalize()?;
+        if !canonical.is_file() {
+            continue;
+        }
+        let rel_entry = canonical
+            .strip_prefix(&base_canonical)
+            .map_err(|_| ProjectError::Symlink(entry, canonical.clone()))?
+            .to_path_buf();
```
colinmarc, line 431: nit: `to_path_buf()` shouldn't be necessary here

src/project.rs
giacomopiccinini: I guess not, it's just something that could happen in principle. I am ok with removing it.

src/project.rs
giacomopiccinini: used below for ` Err(ProjectError::ExtensionNotAllowed(rel_entry))`.

colinmarc: Ok, then you can move the `to_owned()` there :)

tests/fixtures/sql_prohibited_extension/views/age.sql
```diff
@@ -0,0 +1 @@
+SELECT COUNT(*) - COUNT(Age) AS n_nulls_age FROM titanic
```
colinmarc: nit: trailing newlines on these files

src/project.rs
giacomopiccinini: removed

tests/fixtures/sql_prohibited_extension/views/age.sql
giacomopiccinini: done

src/project.rs
```diff
@@ -331,10 +375,166 @@ impl ProjectFile {
...
+    for component in pattern.components() {
+        match component {
+            Component::CurDir => continue,
+            // Patterns involving ".." are not allowrd
+            Component::ParentDir => return Err(ProjectError::GlobPatternNotAllowed(p.to_string())),
+            other => {
+                if !glob.is_empty() {
+                    glob.push('/');
+                }
+                glob.push_str(
```
colinmarc: Why rebuild the glob here? `NonUtf8Pattern` is unreachable since you're starting with a `String` (which has to be utf8).

src/project.rs
```diff
@@ -331,10 +375,166 @@ impl ProjectFile {
...
+        .build()
+    {
+        let entry = entry?.into_path();
+        let canonical = entry.canonicalize()?;
+
+        if !canonical.is_file() {
+            continue;
+        }
+
+        let rel_entry = canonical
```
colinmarc, line 428: You iterate over the same files in both calls to `resolve_includes`. You can just move the whitelist here, and check it in the one `Walk`. no need for two calls.

When you hit `bauplan_project.yaml` here, `rel_entry` will be exactly that, so you can check it with an exact match.

src/project.rs
```diff
@@ -256,6 +276,9 @@ pub struct ProjectFile {
     /// Parameters for models.
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub parameters: BTreeMap<String, ParameterDefault>,
+    /// Additional paths to include as glob expressions.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub include_paths: Option<Vec<String>>,
```
colinmarc: I think with `serde(default)` you can just use `Vec` instead of `Option<Vec>`

src/project.rs
```diff
@@ -302,27 +321,47 @@ impl ProjectFile {
...
+        ];
+
+        // Additional, user-provided patterns.
+        let additional_patterns = self
+            .include_paths
+            .iter()
+            .map(|p| resolve_pattern(p))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let patterns: Vec<&str> = base_patterns
```
colinmarc: nit: this is a bit awkward, I would move the declaration of `base_patterns` into `resolve_includes`.

src/project.rs
```diff
@@ -302,27 +321,47 @@ impl ProjectFile {
...
+            .iter()
+            .copied()
+            .chain(additional_patterns.iter().map(String::as_str))
+            .collect();
+
+        let files: HashSet<PathBuf> = resolve_includes(project_dir, &patterns)?.collect();
```
colinmarc: nit: you don't need to `collect()` this, you just iterate over it below :)

src/project.rs
colinmarc: eg

```rust
for p in patterns {
    glob_builder.add(Glob::new(p.as_ref())?);
}
for p in base_patterns {
    glob_builder.add(Glob::new(p).unwrap());
}
```

src/project.rs
```diff
@@ -331,10 +370,118 @@ impl ProjectFile {
...
+) -> Result<impl Iterator<Item = PathBuf>, ProjectError> {
+    let base_canonical = base.canonicalize()?;
+
+    // Build a glob set, matching all the required patterns.
+    let mut glob_builder = GlobSetBuilder::new();
+
+    for p in patterns {
+        glob_builder.add(Glob::new(p.as_ref())?);
+    }
+    let set = glob_builder.build()?;
```
colinmarc, line 416: nit: the whitespace in this function is kinda wacky. as a general rule I leave a blank line after `}` but group lines together if they're related.

colinmarc (approved):
Just a few nits left, good work!

PR #285, opened 2026-06-08
feat(longbow): bump iroh to 1.0.0

colinmarc:
We also update the interfaces to take a borrow endpoint, so we can reuse it in the pysdk.

In particular, this pulls in
https://github.com/n0-computer/iroh/pull/4270, which makes `endpoint.close()` significantly faster, so we can now gracefully shutdown.


bigluck (approved):
🚀

PR #289, opened 2026-06-10
feat(pysdk): expose job error messages

bigluck:
Surface failed job messages on Job and JobContext so clients can diagnose failures directly. 
Show the message in job get output when one is available.

### DIFF
--- a/src/proto/bpln_proto/commander/service/v2/job_context.proto
+++ b/src/proto/bpln_proto/commander/service/v2/job_context.proto
@@ -63,6 +63,9 @@ message JobContext {
 
   // Logs for Job
   repeated RunnerEvent job_events = 7;
+
+  // Error message for failed jobs, when available.
+  optional string error_message = 11;
 }
 
 message JobError {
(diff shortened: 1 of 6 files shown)


src/proto/bpln_proto/commander/service/v2/job_context.proto
```diff
@@ -63,6 +63,9 @@ message JobContext {
 
   // Logs for Job
   repeated RunnerEvent job_events = 7;
+
+  // Error message for failed jobs, when available.
+  optional string error_message = 11;
```
colinmarc, line 68: `prost` should (as of recently) generate this as `Option<String>` I believe? If not then `optional` is useless and we don't need it.

bigluck: Yes, `prost` generates this as `Option<String>`, so optional is useful here

colinmarc: Then why is `error_message: ctx.error_message.filter(|s| !s.is_empty())` necessary? Shouldn't that already be `None`? Sorry if I'm missing something.

colinmarc (approved):

PR #290, opened 2026-06-10
build: ditch release-plz

colinmarc:
We use it for 2 things:

 - Generating releases and tags
 - Creating a PR with a version bump

Both just use `git cliff` under the hood. Also, the PR workflow is currently failing because it's trying to `cargo package` even though we are in "git only" mode with `publish = false`.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,155 @@
...
+            echo "::error::Failed to check release ${tag}: ${resp}"
+            exit 1
+          fi
+      - name: Create GitHub release
+        if: steps.check.outputs.release_created == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.check.outputs.tag }}
+        run: |
+          notes=$(git-cliff --latest)
```
bigluck: I tested it locally and I think there's something odd here; after the version bump it still emits the latest tagged release notes (`0.1.12` in my test), while `git-cliff --bump --unreleased` emits the new 0.1.13 notes.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,155 @@
...
+          persist-credentials: false
+      - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+      - name: Compute version bump
+        id: bump
+        run: |
+          bump=$(git-cliff --bumped-version)
+          current=v$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[] | select(.name=="bauplan") | .version')
+          echo "current_version=${current}" >> "$GITHUB_OUTPUT"
+          echo "bumped_version=${bump}" >> "$GITHUB_OUTPUT"
+          echo "Current: v${current}, bump: ${bump}"
```
bigluck: nit: `v{current}` being current = `v0...` will produce a log similar to `Current: vv0...`

.github/workflows/release.yaml
```diff
@@ -0,0 +1,155 @@
+name: Release
+
+on:
+  push:
+    branches: [main]
+
+permissions: {}
+
+jobs:
+  # If the version in Cargo.toml doesn't match the latest git tag, create a
```
bigluck: nit: I think this comment is slightly stale: the code checks whether a gh release exists for the current cargo version tag, not whether Cargo.toml matches the latest git tag

bigluck (approved):

.github/workflows/release.yaml
colinmarc: What's odd about that? That matches what I would expect.

colinmarc: You mean that it's checking the git tag?

colinmarc: ...and this is before we actually create the tag. Fixed.

PR #295, opened 2026-06-10
fix(cli): accept api_endpoint in config set

bigluck:
api_endpoint is already resolved from profile config, but clap rejected it as an unsupported setting.

### DIFF
--- a/src/cli/config.rs
+++ b/src/cli/config.rs
@@ -23,6 +23,8 @@ pub(crate) enum ConfigCommand {
 #[clap(rename_all = "snake_case")]
 pub(crate) enum ConfigSetting {
     ApiKey,
+    #[value(hide = true)]
+    ApiEndpoint,
     ActiveBranch,
 }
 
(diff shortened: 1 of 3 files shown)


colinmarc:
I guess this was intended, as `api_endpoint` is hidden/internal, and there are other ways to set it. Is there a specific reason you ran into this?

bigluck:
I agree that it is hidden, as is the active branch, but the active branch can also be set today with config set. The main reason is that during testing on Windows, I found it more convenient to use the CLI than the graphical interface, finding the yaml file and editing it with Notepad.
But I suppose my underlying reason is: if something can be set with a config file, why prevent some configs (hidden, yes) supported by the CLI?

colinmarc:
Maybe we can support it but not expose it in the help text, as a compromise? I think with clap you can hide it. I just don't think users will ever need to know about it, and it's overhead for them.

bigluck:
I think it's an excellent compromise

colinmarc (approved):

PR #297, opened 2026-06-11
fix(cli): add `version` to generated pyproject in init

colinmarc:
This fixes `uv` invocations, which otherwise fail with:

```
error: Failed to parse: `pyproject.toml`
  Caused by: TOML parse error at line 1, column 1
  |
1 | [project]
  | ^^^^^^^^^
`pyproject.toml` is using the `[project]` table, but the required `project.version` field is neither s
```

We also add `ruff` and `ty` as dev dependencies, and `polars`, so that `models.py` passes typechecking. Finally, there's a typechecking error, because `pl.from_arrow` returns `DataFrame | Series`.

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -73,12 +73,22 @@ pub(crate) fn handle(args: InitArgs) -> anyhow::Result<()> {
     //
     // [project]
     // name = "my_project"
-    // requires-python = ">=3.12"
+    // version = "0.1.0"
+    // requires-python = "~=3.12"
     // dependencies = ["bauplan=~x.x.x"]
+    //
+    // [dependency-groups]
+    // dev = [
+    //     "ruff>=0.15",
+    //     "ty>=0.0.48",
+    //     "polars>=1"
+    // ]
     let bauplan_version = concat!("bauplan~=", env!("CARGO_PKG_VERSION")).into();
 
+    let mut pyproject = toml::Table::default();
     let mut project = toml::Table::default();
     project.insert("name".into(), toml::Value::String(project_name.clone()));
+    project.insert("version".into(), toml::Value::String("0.1.0".into()));
     project.insert(
         "requires-python".into(),
         toml::Value::String("~=3.12".into()),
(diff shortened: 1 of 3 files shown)


.github/workflows/release-plz.yaml
```diff
@@ -0,0 +1,86 @@
+name: Release-plz
```
bigluck: ?

colinmarc: 🤷

src/cli/init/models.py
```diff
@@ -20,9 +20,9 @@ def survival_rate_by_age(
     | 19       | 0.3           |
     | 20       | 0.4           |
     """
-    import polars as pl # noqa
+    import polars as pl
```
bigluck, line 23: I get the idea of favoring polars, but this is still a blueprint, and it adds a dev dependency for everyone from the start. I think it's ok for now.

What I'd really like one day is to evolve init into something that:
1. queries a public git repo (something like bauplan-init-blueprints)
2. fetches the list of blueprints from remote and shows them to the user
3. lets the user pick which project to "import/clone" as a starting point
4. downloads the project folder via git and overwrites the project id and name

I'm not asking to do it here of course, but it would be a nice feature (similar to what serverless does for example), and it would let us keep the cli/pysdk separate from the available blueprints.

src/cli/init.rs
bigluck, line 74: `version`?

src/cli/init.rs
```diff
@@ -75,10 +75,19 @@ pub(crate) fn handle(args: InitArgs) -> anyhow::Result<()> {
     // name = "my_project"
     // requires-python = ">=3.12"
```
bigluck: `~=3.12`?

src/cli/init.rs
```diff
@@ -75,10 +75,19 @@ pub(crate) fn handle(args: InitArgs) -> anyhow::Result<()> {
...
+    // ]
     let bauplan_version = concat!("bauplan~=", env!("CARGO_PKG_VERSION")).into();
 
+    let mut pyproject = toml::Table::default();
     let mut project = toml::Table::default();
     project.insert("name".into(), toml::Value::String(project_name.clone()));
+    project.insert("version".into(), toml::Value::String("0.1.0".into()));
     project.insert(
         "requires-python".into(),
         toml::Value::String("~=3.12".into()),
```
bigluck, line 94: isn't `~=3.12` equal to `>=3.12,<4`? Isn't there a risk that when a new version of Python is released, the user will initialize `uv` with a version that is not yet compatible with our `bauplan` library?

bigluck (approved):

src/cli/init/models.py
colinmarc: I know you've asked for this before, but I'm really not a fan... it's way too much infrastructure and overhead for the user. They just need scaffolding and a blank canvas.

bigluck: `too much infrastructure` -> a github repo? :)

src/cli/init.rs
colinmarc: You mean python 4?

bigluck: no I meant 3.16 for which maybe we don't have a plan ready yet

colinmarc: I think `uv` always picks the lower end of the range? Otherwise it would already break, because we don't support 3.14.

bigluck: from what I know if you don't have .python-version uv try to use a version already present in the path, or download the latest version available, and it might download 3.16 if available

PR #306, opened 2026-06-15
build: install cargo-edit only in the release job

bigluck:
The 0.1.13 release failed building the linux/arm64 wheel. cargo-edit was a global mise tool, so mise-action tried to build it on every runner, and the arm64 image ships rustc 1.91, below cargo-edit 0.13.11's 1.92 minimum.

cargo-edit is only used by `cargo set-version` in the release PR job, so I install it there instead of having every build pull it in. The wheel build runners no longer touch it.

Failed run:
- https://github.com/BauplanLabs/bauplan/actions/runs/27538507248/job/81400778836

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -115,6 +115,7 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
       - uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+      - uses: ./.github/actions/setup-caches
       - name: Compute version bump
         id: bump
         run: |
@@ -123,6 +124,8 @@ jobs:
           echo "current_version=${current}" >> "$GITHUB_OUTPUT"
           echo "bumped_version=${bump}" >> "$GITHUB_OUTPUT"
           echo "Current: ${current}, bump: ${bump}"
+      - name: Install cargo-edit
+        run: cargo install cargo-edit@0.13 --locked
       - name: Prepare release
         id: prepare
         if: steps.bump.outputs.current_version != steps.bump.outputs.bumped_version
--- a/mise.toml
+++ b/mise.toml
@@ -1,6 +1,5 @@
 [tools]
 buf = "1"
-"cargo:cargo-edit" = "0.13"
 git-cliff = "2"
 bun = "1.2.18"
 just = "latest"


.github/workflows/publish.yaml
```diff
@@ -12,11 +12,19 @@ on:
         description: "Release tag (e.g. v0.1.2)"
         required: true
         type: string
+  # build (no publish) on PRs that touch the build setup
```
colinmarc: I'm not sure this is worth it. The build jobs are super long and we touch Cargo.toml in almost every PR.

.github/workflows/publish.yaml
```diff
@@ -12,11 +12,19 @@ on:
...
+  pull_request:
+    paths:
+      - mise.toml
+      - Cargo.toml
+      - pyproject.toml
+      - build.rs
```
colinmarc: If we do keep this, we should remove build.rs.

mise.toml
```diff
@@ -12,3 +12,7 @@ rust = { version = "1.95.0", profile = "default" }
 uv = "0.10"
 vale = "3.14.1"
 zizmor = "1.23.1"
+
+[env]
+# build cargo: tools with the pinned rust, not the runner's default (mise#6795)
```
colinmarc: Rather than this fix, let's remove `cargo-edit` from `mise.toml`, and add `cargo install cargo-edit` only to the PR job.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ tracing = { version = "0.1" }
 url = "2"
 
 [workspace.package]
-version = "0.1.13"
+version = "0.1.12"
```
colinmarc: You don't need to decrement the version. It'll also probably break all the tooling.

.github/workflows/publish.yaml
bigluck: This is how I see it: it's not perfect now because it does it for every PR, and I agree with you, but right now I needed it to validate that the fix works, and it works:

https://github.com/BauplanLabs/bauplan/actions/runs/27544027584

But I wouldn't want this job to be launched only upon in main because then it happens like today when we merge the PR and it fails there.

So in my opinion it should be kept at least as a pre-merge approval for a new release.

mise.toml
bigluck: Don't we risk shifting the problem a bit? Because if we then do `cargo install cargo-edit,` don't we risk having a decupling between our versions and the one in the cicd?

.github/workflows/publish.yaml
colinmarc: That's even worse, because then we'll run it literally on every commit (when the PR gets rebased). The job takes a long time and zizmor doesn't allow us to enable caching because it could be poisoned.

I don't really mind hitting these issues every once in a while, they're quickly fixed.

mise.toml
colinmarc: No, we don't use cargo-edit locally at all, so there's no "our version". Also, it's really simple what it's doing, and the version shouldn't matter.

Cargo.toml
bigluck: I know we did this a while ago, and it was for the 0.1.11 release, but if you look in the changelog file that release is completely missing, so something went wrong.

colinmarc: I think that's because it was a release with no changelog-relevant fixes. In any case nothing would've deleted the changelog entries, because there was no revert.

colinmarc: https://github.com/BauplanLabs/bauplan/commit/8c4aa28f1beaa33c6438ace56535deb1457a5c6d

.github/workflows/release.yaml
```diff
@@ -129,6 +129,9 @@ jobs:
         env:
           VERSION: ${{ steps.bump.outputs.bumped_version }}
         run: |
+          # cargo-edit provides `cargo set-version`; install it just for this job.
```
colinmarc: nit: can you add this as a separate step?

Also we should make sure the rust build is cached.

colinmarc (approved):

.github/workflows/release.yaml
```diff
@@ -123,6 +123,16 @@ jobs:
           echo "current_version=${current}" >> "$GITHUB_OUTPUT"
           echo "bumped_version=${bump}" >> "$GITHUB_OUTPUT"
           echo "Current: ${current}, bump: ${bump}"
+      # cargo-edit provides `cargo set-version`; build it once and cache it.
```
colinmarc: comment is unecessary.

.github/workflows/release.yaml
```diff
@@ -123,6 +123,16 @@ jobs:
...
+      - name: Cache cargo-edit
+        id: cargo-edit-cache
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: ~/.cargo/bin/cargo-set-version
```
colinmarc: just save the cargo cache (you can copy/paste from all the other workflows)

PR #321, opened 2026-06-17
build: pass --unreleased to git cliff

colinmarc:
In local testing, `--bumped-version` without `--unreleased` is a noop, because it can't "see" the commits that would make up the bumped version. This seems like questionable UX, but what do I know.


bigluck (approved):

bigluck:
Merci

PR #325, opened 2026-06-18
fix(cli): align checkout -b tip and help with real behavior

### LINKED ISSUE #241: `bauplan checkout -b` fails when branch exists, contradicting documented `--if-not-exists` behavior
**Title:** `bauplan checkout -b` fails when branch exists, contradicting documented `--if-not-exists` behavior

---

### Summary

There are two related issues with `bauplan checkout -b`:

1. **Behavior does not match documentation.** The `-b` flag is documented as equivalent to `branch create --if-not-exists`, but the command fails with `BRANCH_EXISTS` when the branch already exists.
2. **Confusing tip after `branch create`.** After successfully creating a branch via `bauplan branch create <name>`, the CLI suggests running `bauplan checkout -b "<name>"` to create and switch in one command. Following that tip immediately fails because the branch was just created.

### Issue 1: `-b` does not honor `--if-not-exists` semantics

From `bauplan help`:

```
-b
    Create the branch first (equivalent to "branch create --if-not-exists")
```

The `--if-not-exists` semantics imply the command should succeed silently (and check out the existing branch) when the branch already exists. Instead, it errors:

```
$ bauplan checkout -b mapav19.mauro
Error: Failed to create branch

Caused by:
    BRANCH_EXISTS: Branch "mapav19.mauro" already exists
```

### Issue 2: Misleading post-create tip

Reproduction:

```
$ bauplan branch create mapav19.mauro
Created branch "mapav19.mauro"
TIP: To create and switch to a branch in one command, run:
        bauplan checkout -b "mapav19.mauro"

$ bauplan checkout -b "mapav19.mauro"
Error: Failed to create branch

Caused by:
    BRANCH_EXISTS: Branch "mapav19.mauro" already exists
```

The tip is shown immediately after the branch is created, so the suggested command is guaranteed to fail in that exact context. This is confusing for users who follow the tip literally.

### Expected behavior

- **Issue 1:** `bauplan checkout -b <name>` should either (a) check out the existing branch when it already exists (matching `--if-not-exists` semantics), or (b) the help text should be updated to reflect the actual behavior.
- **Issue 2:** The tip shown after `bauplan branch create` should either be removed in that context, or rephrased so it does not suggest a command that will fail immediately afterward (e.g., suggest `bauplan checkout <name>` instead, since the branch already exists at that point).

### Environment

- bauplan 0.1.9 (HEAD:928733e, release, macos/aarch64)

bigluck:
This fixes the two problems reported in #241.

The tip printed after `bauplan branch create` suggested running `bauplan checkout -b "<name>"`, but the branch had just been created, so that command always failed with `BRANCH_EXISTS`. I changed the tip to suggest plain `bauplan checkout "<name>"`, which just switches to the branch that already exists.

I also fixed the help text for the `-b` flag. It claimed to be equivalent to `branch create --if-not-exists`, but the command actually fails when the branch already exists. The help now describes the real behavior instead.

I did not change the behavior of `-b` itself: failing when the branch already exists is the intended, git-like behavior.

Closes #241

### DIFF
--- a/src/cli/branch.rs
+++ b/src/cli/branch.rs
@@ -334,9 +334,9 @@ fn create_branch(cli: &Cli, args: BranchCreateArgs) -> anyhow::Result<()> {
 
     eprintln!("Created branch \"{branch_name}\"");
     anstream::eprintln!(
-        "{GREEN}TIP:{GREEN:#} To create and switch to a branch in one command, run:",
+        "{GREEN}TIP:{GREEN:#} To switch to the new branch, run:",
     );
-    eprintln!("\tbauplan checkout -b {branch_name:?}");
+    eprintln!("\tbauplan checkout {branch_name:?}");
     Ok(())
 }
 
--- a/src/cli/checkout.rs
+++ b/src/cli/checkout.rs
@@ -19,7 +19,7 @@ use crate::cli::{Cli, color::CliExamples, yaml};
 pub(crate) struct CheckoutArgs {
     /// Branch name
     pub branch_name: String,
-    /// Create the branch first (equivalent to "branch create --if-not-exists")
+    /// Create the branch before switching (fails if it already exists)
     #[arg(short = 'b')]
     pub create: bool,
     /// Ref from which to create when using -b. If not specified,


colinmarc (approved):
Thank you!

PR #327, opened 2026-06-18
test: remove fixtures which use intermediate SQL models

colinmarc:
Support for this (currently done via SQLGlot and the old runtime) is going away with the new infrastructure. This removes the tests that depend on that behavior in advance.


bigluck (approved):
LGTM, thanks

PR #337, opened 2026-06-25
fix(cli): amend glob pattern construction

giacomopiccinini:
The current implementation in 0.1.14 introduced a bug. 

When doing `Glob::new("*.sql")` for base (i.e., not user defined) patterns, Glob will automatically match any `.sql` file, regardless of it being in the base project directory or in a nested one. This was not present in the earlier versions because:
1. We used a different globbing strategy overall; 
2. The dir walking was limited to depth 1, so there was no way nested files could end up being included. 

You can reproduce with e.g. this tree
```
.
├── bauplan_project.yaml
├── models.py
├── my_file.sql
├── pyproject.toml
└── views
    └── my_view.sql
```
where `models.py` reads
```
import bauplan


@bauplan.python("3.12", pip={"polars": "1.37"})
@bauplan.model()
def survival_rate_by_age(
    passengers=bauplan.Model(
        "titanic", columns=["Age", "Survived"], filter="Age IS NOT NULL"
    ),
):
    """
    Bins passengers by age and returns survival rate per bin.

    Returned table:
    | Age | survival_rate |
    |----------|---------------|
    | 0        | 1.0           |
    | ...      | ...           |
    | 19       | 0.3           |
    | 20       | 0.4           |
    """
    import polars as pl

    df = pl.DataFrame(passengers)
    return (
        df.group_by(pl.col("Age").floor())
        .agg(pl.col("Survived").mean().alias("survival_rate"))
        .sort("Age")
        .to_arrow()
    )


@bauplan.model()
@bauplan.python()
def dumb(data=bauplan.Model("my_file")):
    print("hey")
    return data
```
and `bauplan_project.yaml` does not contain any user-defined paths. In Bauplan, this produces the following plan:
```
=> DAG
=>  ----------------------
=> | survival_rate_by_age |
=>  ----------------------
=>  ----------------------
=> |       my_file        |
=>  ----------------------
=>   |
=>   |
=>   v
=>  ----------------------
=> |         dumb         |
=>  ----------------------
=>  ----------------------
=> |       my_view        |
=>  ----------------------
```

The solution is to replace _both_ for Bauplan- and user-defined patterns

`Glob::new(pattern)` -> `GlobBuilder::new(pattern).literal_separator(true).build()`

[Doc on literal_separator](https://docs.rs/globset/latest/globset/struct.GlobBuilder.html#method.literal_separator)

I have added a few fixtures to test this, and it solves the issue. 

## Additional testing

I have also tried to: 

1. Keep the old `Glob::new` pattern and replace the base patterns with e.g. `./*.sql` but it still produces the error; 
2. Keep the old `Glob::new` on user-defined patterns only, but it fails when e.g. user includes `views/*.sql` but Bauplan also reads from e.g. `views/nested/*.sql`.


tests/cli/run.rs
```diff
@@ -537,3 +537,35 @@ fn sql_prohibited_extension() {
         .failure()
         .stderr(contains("glob pattern not allowed"));
 }
+
+#[test]
```
bigluck: nit: I would add also a positive test, like `views/*.sql` that does match `views/age.sql` at the first level. Right now `sql_nested` mixes `views/*.sql` and `views/**/*.sql`, so from a success you cannot tell if `views/*.sql` matched anything; the ** could be doing all the work

tests/cli/run.rs
```diff
@@ -537,3 +537,35 @@ fn sql_prohibited_extension() {
...
+            "-p",
+            "tests/fixtures/sql_nested_not_included",
+        ])
+        .assert()
+        .failure()
+        .stderr(contains("not found on branch"));
+}
+
+#[test]
+fn sql_included_first_level_only() {
```
bigluck: nit: the name `sql_included_first_level_only` is a bit confusing, the file here is at `views/nested/age.sql` and the test checks that it is not found. Maybe a name like `sql_user_pattern_not_recursive`?

bigluck (approved):
merci

tests/cli/run.rs
giacomopiccinini: true, thanks, will do.

tests/cli/run.rs
giacomopiccinini: makes sense

tests/cli/run.rs
```diff
@@ -537,3 +537,65 @@ fn sql_prohibited_extension() {
...
+#[test]
+fn sql_nested_not_included() {
+    bauplan()
+        .args([
+            "run",
+            "--dry-run",
+            "--cache",
+            "off",
+            "-p",
+            "tests/fixtures/sql_nested_not_included",
```
colinmarc: I think it's probably better to construct a unit test for this behavior (in `project.rs`) instead of adding a whole test fixture. You can create a bunch of files in a `tempdir` and run `resolve_includes` on it. That test will have the same value but be significantly faster.

That goes for all the tests here.

giacomopiccinini: Thanks, done.

src/project.rs
colinmarc: Maybe for the base case that there are no include paths we can set max_depth 1? It's only an optimization but probably worthwhile.

colinmarc (approved):

src/project.rs
```diff
@@ -484,4 +488,93 @@ mod tests {
...
+            Err(ProjectError::GlobPatternNotAllowed(..))
+        ));
+    }
+
+    #[test]
+    fn resolve_includes_includes_top_level_sql() {
+        let tmp = tempfile::tempdir().unwrap();
+        let proj = tmp.path().join("proj");
+        std::fs::create_dir_all(&proj).unwrap();
+        std::fs::write(proj.join("age.sql"), "SELECT 1").unwrap();
```
colinmarc: nit: if you change the test signature to return anyhow::Result, you can delete all these unwraps

src/project.rs
giacomopiccinini: fixed.

src/project.rs
giacomopiccinini: fixed.

PR #342, opened 2026-06-30
Longbow client tweaks

colinmarc:
This fixes two issues:

 - If tasks are skipped and no TaskStartedEvent is sent, the tasks wouldn't display as spinners at all in the CLI.
 - If the job ended before the tasks' output was drained, we wouldn't get a chance to see it.

### DIFF
--- a/src/cli/run.rs
+++ b/src/cli/run.rs
@@ -222,6 +222,7 @@ pub(crate) async fn monitor_job_progress(
     let stream = client.monitor_job(monitor_req, Arc::clone(&endpoint));
     futures::pin_mut!(stream);
 
+    let mut outcome = None;
     loop {
         let res = tokio::select! {
             v = stream.try_next() => v,
@@ -234,7 +235,7 @@ pub(crate) async fn monitor_job_progress(
 
         let event = match res {
             Ok(Some(v)) => v,
-            Ok(None) => bail!("no JobCompletion event found"),
+            Ok(None) => break,
             Err(ref e)
                 if e.code() == tonic::Code::Cancelled
                     || e.code() == tonic::Code::DeadlineExceeded =>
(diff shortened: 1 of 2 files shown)


src/cli/run.rs
bigluck, line 265: nit: we close the endpoint and return on JobCompletion, but my understanding is that the longbow drain only runs when the grpc stream ends. Since the server sends JobCompletion then closes, does that drain ever run here?

bigluck (approved):

src/cli/run.rs
colinmarc: That's not a nit, that's a really good catch. Fixed.

src/grpc/job.rs
```diff
@@ -408,9 +410,30 @@ impl futures::Stream for JobEventStream {
...
+
+            if let Some(inner) = this.inner.as_mut()
+                && let Poll::Ready(v) = Pin::new(inner).poll_next(cx)
+            {
                 let Some(resp) = v else {
-                    return Poll::Ready(None);
+                    // The stream has ended. If there are still longbow
+                    // connection tasks, wait for those. Otherwise, we could
+                    // exit before printing all the output for those tasks.
+                    this.inner.take();
```
bigluck, line 441: on inner EOF, if the JoinSet is already empty, my understanding is that we return Ready(None) before draining longbow_output_rx, so buffered longbow lines get lost

colinmarc: The JoinSet can't be empty until all the longbow tasks writing to that channel finish. Additionally, the loop is "biased" towards the channel, so we always read from it first.

colinmarc: Oh, I biased it in an earlier version but didn't now. I'll swap the order.

PR #345, opened 2026-07-02
fix(longbow): keep the connection alive while the task is quiet

bigluck:
Iroh's default keepalive lines up with the 5s idle timeout, so when user code goes quiet we drop the log connection; it shows up mostly on slow
relay-only paths.
I fixed it by sending keepalives more often to stay attached.

### DIFF
--- a/bauplan-longbow/src/lib.rs
+++ b/bauplan-longbow/src/lib.rs
@@ -117,6 +117,14 @@ impl iroh::endpoint::presets::Preset for BauplanPreset {
             .alpns(vec![ALPN.to_owned()])
             .relay_mode(RelayMode::Custom(self.relay_map()))
             .crypto_provider(std::sync::Arc::new(rustls::crypto::ring::default_provider()))
+            // Drop idle connections after 5s instead of iroh's 30s default, and
+            // keep-alive more often than iroh's 5s default so a quiet one stays up.
+            .transport_config(
+                iroh::endpoint::QuicTransportConfig::builder()
+                    .max_idle_timeout(Some(std::time::Duration::from_secs(5).try_into().unwrap()))
+                    .keep_alive_interval(std::time::Duration::from_secs(1))
+                    .build(),
+            )
     }
 }
 


bauplan-longbow/src/client.rs
```diff
@@ -96,8 +99,15 @@ async fn connect(
 /// Returns a handle with a separate stream for stdout and stderr, respectively.
 /// Dropping the handle closes the connection.
 pub async fn attach_task(endpoint: &Endpoint, server: EndpointAddr) -> Result<AttachedTask, Error> {
+    // iroh's default timeout is 5s, so if the task stops producing logs for
```
colinmarc: The idle timeout is 30s by default, we just set it to 5s on the backend. (We should do the same here).

bauplan-longbow/src/client.rs
```diff
@@ -96,8 +99,15 @@ async fn connect(
...
+    // that long we lose the connection. Ping more often to keep it alive.
+    let transport = QuicTransportConfig::builder()
+        .keep_alive_interval(Duration::from_secs(2))
```
colinmarc:
```suggestion
        .keep_alive_interval(Duration::from_secs(1))
```

colinmarc (approved):
Can we do this globally in the `Preset`?

bauplan-longbow/src/lib.rs
```diff
@@ -117,6 +117,14 @@ impl iroh::endpoint::presets::Preset for BauplanPreset {
             .alpns(vec![ALPN.to_owned()])
             .relay_mode(RelayMode::Custom(self.relay_map()))
             .crypto_provider(std::sync::Arc::new(rustls::crypto::ring::default_provider()))
+            // Match the backend's 5s idle timeout (default is 30s) and keep-alive
```
colinmarc: nit: this preset is applied to the backend, as well.

colinmarc (approved):

colinmarc:
Thank you!

PR #346, opened 2026-07-02
fix(cli): trust the bundled mozilla roots so tls works on fresh windows

bigluck:
On fresh Windows machines every gRPC call failed with `invalid peer certificate: UnknownIssuer`.

Windows does not preinstall all root CAs; it downloads them on demand (including `Amazon Root CA 1`, which signs our ACM certificates), but only when the validation goes through the native CryptoAPI.

Our HTTP calls were already safe because ureq uses the platform verifier; the tonic channels instead use `with_native_roots()`, which only takes a snapshot of the Windows store and never triggers that download, so on a fresh machine the Amazon root is simply not there.

The fix adds the Mozilla bundle (`webpki-roots`) on top of the native roots for the two tonic channels (commander and flight): the bundle ships the Amazon roots, so the handshake works even when the store was never populated. We keep the native roots too, so corporate proxies with custom CAs keep working.

The bundle is gated to Windows (`cfg` plus a per-target feature), so linux and mac keep exactly the same trust store as before.

### DIFF
--- a/src/grpc.rs
+++ b/src/grpc.rs
@@ -46,7 +46,7 @@ impl Client {
     ) -> Result<Self, tonic::transport::Error> {
         let api_endpoint = profile.api_endpoint.clone();
         let channel = Channel::builder(api_endpoint)
-            .tls_config(ClientTlsConfig::new().with_native_roots())?
+            .tls_config(ClientTlsConfig::new().with_enabled_roots())?
             .timeout(timeout)
             .user_agent(&profile.user_agent)?
             .connect_lazy();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5998,6 +5998,7 @@ dependencies = [
  "tower-layer",
  "tower-service",
  "tracing",
+ "webpki-roots",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


src/grpc.rs
```diff
@@ -35,6 +35,16 @@ use crate::{
 };
 use generated::v2_commander_service_client::V2CommanderServiceClient;
 
+/// TLS config for tonic channels. Besides the OS trust store, this trusts the
+/// bundled Mozilla roots: Windows installs many roots (including Amazon's) on
+/// demand via CryptoAPI, which rustls-native-certs never invokes, so a fresh
+/// machine may not have the CA for our endpoints in its store yet.
+pub(crate) fn tls_config() -> ClientTlsConfig {
```
colinmarc: Why is this a function?

src/grpc.rs
```diff
@@ -46,7 +56,7 @@ impl Client {
     ) -> Result<Self, tonic::transport::Error> {
         let api_endpoint = profile.api_endpoint.clone();
         let channel = Channel::builder(api_endpoint)
-            .tls_config(ClientTlsConfig::new().with_native_roots())?
+            .tls_config(tls_config())?
```
colinmarc: `#[cfg(target_os = "windows")]`

colinmarc (approved):

colinmarc (approved):

PR #347, opened 2026-07-06
build: allow specifying version bumps manually

colinmarc:
This lets us trigger the release flow (to create a PR) with a specific version. If we merge a pre-release bump, it also makes the github release a prerelease.


bigluck (approved):
🚀

PR #376, opened 2026-07-28
docs: update expectations page

innazh:
## docs: update expectations page

- removed some non-existing expectations: `expect_column_values_to_be_between`,
`expect_column_mean_to_be_between`, `expect_column_values_to_be_unique`, and
`expect_column_to_exist`
- updated the function names that were wrong: `expect_column_all_nulls` to
`expect_column_all_null`.
- added missing expectations: `expect_column_some_null`,
`expect_column_accepted_values`, `expect_column_not_unique`,
`expect_column_equal_concatenation`.
- added a `## Custom expectations` section, defining the previously undefined
`check_data_recency` helper.

Note: the test are currently failing but it's unrelated to this PR.

### DIFF
--- a/docs/pages/concepts/expectations.mdx
+++ b/docs/pages/concepts/expectations.mdx
@@ -71,14 +71,14 @@ def test_critical_field_completeness(
 
 Bauplan provides a built-in library covering common validation scenarios:
 
-- **Nullness checks**: `expect_column_no_nulls`, `expect_column_all_nulls`
-- **Value constraints**: `expect_column_values_to_be_between`, `expect_column_mean_to_be_between`
-- **Uniqueness**: `expect_column_values_to_be_unique`
-- **Existence**: `expect_column_to_exist`
+- **Nullness checks**: `expect_column_no_nulls`, `expect_column_all_null`, `expect_column_some_null`
+- **Accepted values**: `expect_column_accepted_values`
+- **Uniqueness**: `expect_column_all_unique`, `expect_column_not_unique`
 - **Statistical properties**: `expect_column_mean_greater_than`, `expect_column_mean_greater_or_equal_than`, `expect_column_mean_smaller_than`,
   `expect_column_mean_smaller_or_equal_than`
+- **Derived columns**: `expect_column_equal_concatenation`
 
-You can also write custom expectations or integrate other libraries like [Great Expectations](https://github.com/great-expectations/great_expectations).
+You can also write custom expectations. See [Custom expectations](#custom-expectations) below.
 
 ## Nullness checks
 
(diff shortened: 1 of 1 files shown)


docs/pages/concepts/expectations.mdx
```diff
@@ -171,6 +171,39 @@ expect_column_all_unique(table, 'transaction_id')
...
+#! def check_data_recency(table: pa.Table, column: str, hours: int) -> bool: return True
+import bauplan
+
+@bauplan.expectation()
+@bauplan.python('3.11')
+def test_data_freshness(data=bauplan.Model('model', columns=['timestamp'])):
+    return check_data_recency(data, 'timestamp', hours=24)
+```
+
 ## Failure handling strategies
```
giacomopiccinini, line 207: Not your fault (you didn't modify it in this PR) but I personally dislike the hard vs soft failure and the use of assert. 

My POV here is:
1. Unless there's a bug in the code, expectations should only return `True` or `False`, and not raise an error;
2. Whether a `True` halts the pipeline or not depends on `strict` being on/off, not on an assertion;
3. In any case, any print statement, whether it comes from an `assert` or an explicit `print`, seems for the most part redundant to me. The expectation should cover _just one_ case: if `test_data_freshness` expectation fails, data is obviously stale, no need to log the statement too. 

Cc'ing @bigluck as this is a more philosophical question

bigluck: Apologies for the long reply... oru `examples/learn/02-data-quality-expectations/pipeline/expectations.py` and its README adds further confusion in this regard.

spoiler alert: I'm bringing the discussion to a meta-pseudo-philosophical level.. 

I have the feeling there is an underlying friction between how a python user expects a test to work (`pytest`, that does not consume the return but expects to receive Exceptions for everything, because assert raises an Exception) and how bauplan treats an expectation.

In my opinion, it is legitimate to throw an exception: without type contracts, for example, you might want to check that a certain column does not have null values, but if the column does not exist, that is a good example of a valid exception, and also a good example of a hard stop for the dag and it's not technically a bug in the code but a bug in the shape of the data -> if our runtime treats an exception as a soft failure we should fix it there.
Another example is the user that imports numpy but invokes an invalid function name, that's another hard failure and we should shop the dag -> this is a real bug in the user code.

In other words: I believe an Exception should always stop the DAG.

Then there's a second level of verification on the quality of the data itself: someone produced some dirty data in dev but in the meantime you still want to move forward with pipeline development, i like the idea you can disable strict mode while developing and before the fix arrives upstream.

Which is why I think strict should default to True, with an explicit way to turn it off while developing. It's a breaking change though, and we need to plan it very well.

I also think this `test_*` function (`test_data_freshness`) can be misinterpreted by those who read the code thinking that internally we follow the logic of pytest.. and at the same time `expect_` might be implied by a python user, reading the name of the function, that it throws an exception, but I can also see how this expect comes from libraries like `Great Expectations`, and so I fear we will have to live with my internal conflict.

Ultimately, I agree that printing on success doesn't add value to the user, and I would avoid it if possible. Conversely, I believe it's useful for the user to print something on error, hoping one day we enable users to inspect and query in-memory artifacts for more accurate debugging than a print, which risks dumping the entire table.

innazh: Ok, so from what I'm reading it sounds like an intermediary mediation to fix this page could be:
- Replacing the text in `Failure handling strategies` with a section that describes what happens in case of a failure and how users can use the `strict` flag to change its behaviour. 

How's that ?

In case of the printing - sure, it's just an example func, there could be any text there at all that could be useful to the user in one case or another, I can just remove it bcs it doesn't seem necessary to communicate the point.

giacomopiccinini (approved):
All good, thanks for this and apologies for derailing the conversation.

PR #377, opened 2026-07-28
docs: pin ruff version <0.16

innazh:
## Pin ruff version to below 0.16

ruff 0.16 breaks our CI tests for the docs because of the [B008](https://docs.astral.sh/ruff/rules/function-call-in-default-argument/) rule set by default, pinning a lower version to prevent the problem. This fixes docs CI pipeline.

### DIFF
--- a/src/cli/init.rs
+++ b/src/cli/init.rs
@@ -80,7 +80,7 @@ pub(crate) fn handle(args: InitArgs) -> anyhow::Result<()> {
     //
     // [dependency-groups]
     // dev = [
-    //     "ruff>=0.15",
+    //     "ruff>=0.15,<0.16",
     //     "ty>=0.0.48",
     //     "polars>=1"
     // ]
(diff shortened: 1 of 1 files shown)


src/cli/init.rs
```diff
@@ -103,7 +103,9 @@ pub(crate) fn handle(args: InitArgs) -> anyhow::Result<()> {
     dependency_groups.insert(
         "dev".into(),
         toml::Value::Array(vec![
-            toml::Value::String("ruff>=0.15".into()),
+            // Capped below 0.16: it enables B008 by default, which flags the
+            // `bauplan.Model(...)` argument defaults that models require.
```
colinmarc:
```suggestion
            // ruff 0.16 enables B008 by default, which errors when we use the default argument position for `bauplan.Model()`.
```

colinmarc: (split over multiple lines)

colinmarc (approved):

PR #378, opened 2026-08-04
fix(pysdk): describe query results with the schema of the batches

bigluck:
The flight server advertises a schema that does not always match the batches it sends, so any query returning a list column failed to build the table.


bigluck:
we'll fix it server side

PR #396, opened 2026-08-12
docs: fix video thumbnail, admonitions on brand, surface agent docs earlier

innazh:
Mattia made some comments and suggestions on a call today, so implementing some of the stuff he mentioned and a couple of small fixes I noticed:

- One of the videos on the home page had a grey thumbnail. VideoCard now falls back to the `hqdefault` thumbnail when `maxresdefault` is missing.
- Admonitions used Docusaurus' default green and yellow; success and warning now use the brand blue and yellow.
- Moved the Python 3.11 requirement to the prerequisites list, and used the space for a collapsible tip linking to Bauplan Skills and the agent overview.
- Moved the Agents category above Platform Overview in the sidebar, and dropped the stray `$` from the skill-installer command so it can be copy-pasted.

### DIFF
--- a/docs/src/theme/components/VideoCard.js
+++ b/docs/src/theme/components/VideoCard.js
@@ -1,7 +1,9 @@
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import { Play } from "lucide-react";
 
 const PLAYER_ORIGIN = "https://www.youtube-nocookie.com";
+
+const thumbnailUrl = (id, size) => `https://i.ytimg.com/vi/${id}/${size}.jpg`;
 
 export function VideoCardGrid({ children }) {
   return <div className="grid md:grid-cols-3 gap-4 mb-6">{children}</div>;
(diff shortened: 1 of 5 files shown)


docs/src/theme/components/VideoCard.js
```diff
@@ -42,10 +51,10 @@ export function VideoCard({ id, title, duration, blurb, start }) {
...
           >
             <img
-              src={`https://i.ytimg.com/vi/${id}/maxresdefault.jpg`}
-              onError={(e) => {
-                e.currentTarget.src = `https://i.ytimg.com/vi/${id}/mqdefault.jpg`;
-              }}
+              ref={rejectPlaceholder}
+              src={thumbnail}
+              onLoad={(e) => rejectPlaceholder(e.currentTarget)}
+              onError={() => setThumbnail(thumbnailUrl(id, "hqdefault"))}
```
drin, line 60: is there a reason the rejectPlaceholder isn't used here? it seems better than in `ref` but i don't have much context on fallbacks and error handlers for web stuff

drin (approved):
seems good! i left one comment that's a question and indirect suggestion, but I assume it's not necessary

docs/src/theme/components/VideoCard.js
innazh: These are slightly different cases: `rejectPlaceholder ` rejects the placeholder when it's below a certain width, and since `onError` fires in cases of 404, we just set the thumbnail to the default lower res one - `hqdefault`. I guess technically we could just put `rejectPlaceholder` there since an absent picture is 0, so the check would pass but it's an unnecessary check in this case. Ref handles the case where the browser had fetched the image before React hydrated, so we put `rejectPlaceholder` there to make sure that we didn't load youtube's default grey placeholder.

PR #399, opened 2026-08-14
build: adding dev workflow

drin:
This adds a dev CI action to push dev builds of `bauplan` to pypi

### DIFF
--- a/.github/workflows/dev.yaml
+++ b/.github/workflows/dev.yaml
@@ -1,0 +1,251 @@
+name: Dev build
+
+# Builds and publishes an unreleased dev version of the wheels and CLI from a
+# branch, without tagging, changing the changelog, or deploying docs. Intended
+# for validating other services against unmerged work.
+#
+# Nothing here writes to the repo: the version bump is applied to the checkout
+# in the runner and thrown away.
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch to build (e.g. the head branch of a PR)"
+        required: true
+        type: string
+      version:
+        description: "Override version (Cargo semver, e.g. 0.2.0-dev.7)"
+        required: false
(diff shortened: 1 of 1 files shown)


bigluck (changes requested):
Ciao man, I see what you want to get here, but there are a few technical and security problems that I would like to talk about:

1. as it is now, this workflow can not work; on pypi we have only one trusted publisher and it points to `publish.yaml` with environment `release` (which is a protected gh environment and it can only be invoked from `main`), while here the job that publishes lives inside `dev.yaml` (the new file), so the workflow claim in the `OIDC` token does not match, and pypi rejects it.
2. the build is copied from `publish.yaml`, which is already a `workflow_call` and does the same things. I don't want two pipelines to keep in sync, and they are already drifting (`publish.yaml` uses python 3.10, here you are using 3.11).
3. to make it work we would need to add a second publisher on pypi and before that I would like to understand how we protect it.
4. I would keep `cancel-in-progress: false` as well... `publish.yaml` uses `false` on purpose, because uv publish uploads the wheels one file at a time so a run cancelled in the middle can leave a version on PyPI with only some platforms, and pypi never lets you re-upload a file you already pushed.

For these and other reasons I would prefer not to have a new pipeline dedicated to `dev` builds. I agree it can be useful to have one, but it also brings an additional problem that we cannot ignore: we need to clean up the old dev versions from pypi, otherwise we saturate our space there. Today the project is already consuming 7.1 of 10GB that pypi gives us by and every version we publish is ~600 MB, so after 4 or 5 dev builds we hit the limit and the real releases stop working too. So if we really need this, I would like us to look with more attention at how to do it in the safest way.

drin:
sounds good! this was just an idea that i'm not too attached to, so happy to discuss alternatives

PR #401, opened 2026-08-17
[Quickstart] Remove comments from codeblocks, reorganize section ordering, small language cleanup 

omerbensaadon:
A few changes: 

1. It's common for developers hit the "copy" button on codeblocks. Removing the comments so the commands run effectively
2. Edited some of the language around data exploration
3. Simplified language surrounding data branch (easily maps to git concepts without the explicit explanation)

Maintainers, please feel free to make changes to any or all of this to make it acceptable to merge or simply close the PR! Thanks

### DIFF
--- a/docs/pages/tutorial/01-quick-start.mdx
+++ b/docs/pages/tutorial/01-quick-start.mdx
@@ -18,28 +18,28 @@ other ways to connect an assistant to your lakehouse.
 
 ## Create a branch
 
-Create an isolated [data branch](/concepts/git-for-data/data-branches) to work in - like a git branch, but for your lakehouse:
+Create a new [data branch](/concepts/git-for-data/data-branches) and switch to it:
 
 ```sh
-bauplan checkout -b <YOUR_USERNAME>.quickstart # create a new branch, and switch to it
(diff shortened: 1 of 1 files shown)


omerbensaadon:
Screenshot 2026-08-17 at 9 15 38 AM

Sorry for the noob question but: 
Is init only needed for the tutorial which is why it comes pre-configured to read from the titanic table? Would users expect to need to init for any other reason...?

innazh:
> Screenshot 2026-08-17 at 9 15 38 AM
> Sorry for the noob question but: Is init only needed for the tutorial which is why it comes pre-configured to read from the titanic table? Would users expect to need to init for any other reason...?

hi Omer! First of all, thanks for your PR. 
It's a great question. I think the main purpose of `bauplan init` is the tutorial. It's also a whole project a user can setup in one command, and play around with changing values. I don't think there's any reason a user will specifically need the command but they may want to use it to have some quick boilerplate ready that they can edit and make their own.

Overall the PR looks good to me but I'll hold off before approving/merging to see if you have any additional questions or input.

innazh:
Moved to: https://github.com/BauplanLabs/bauplan/pull/409.

Issue #398, opened 2026-08-14
Job listing inconsistencies


giacomopiccinini:
@colinmarc I have noticed the following inconsistencies when using `client.get_jobs` from the SDK. Happy to take on some of these myself :) 

1. **Filter by Job Kind when Unspecified** 

Bauplan admits a `bauplan.JobKind.UNSPECIFIED` job kind, but these cannot be filtered upon. That is, 
```python
client.get_jobs(
    filter_by_kinds=bauplan.JobKind.UNSPECIFIED,
)
```
won't work. As in, it won't fail but won't return jobs whose kind is unspecified either. 

2. **Minor inconsistency of bauplan.JobKind.UNSPECIFIED**

If you do `str(bauplan.JobKind.UNSPECIFIED)` you get back `Unknown` instead of `Unspecified`. Not a big deal, but it's not consistent. 

3. **Create external table jobs listed as unspecified**

Commands stemming from `client.create_external_table_from_parquet` are saved with kind `bauplan.JobKind.UNSPECIFIED`

4. **Filter by Job Kind has legacy options**

From the CLI (`0.2.0-rc.1`) we get that `bauplan job ls --help` produces, among other things,

```sh
-k, --kind <KIND>
          Filter by job kind

          [possible values: run, query, import-plan-create, import-plan-apply, table-plan-create, table-plan-apply, table-import]
```
As remarked above, the external tables are not part of the admissible kinds. Moreover, the `import-plan-create, import-plan-apply` are likely legacy kinds.

PR #403, opened 2026-08-18
feat: add external-table-create job kind, fix unspecified name

bigluck:
Aligning the client job kinds with the commander protobuf definitions, see:
- https://github.com/BauplanLabs/bauplan/issues/398


src/grpc/job.rs
```diff
@@ -130,19 +130,21 @@ pub enum JobKind {
...
     TablePlanCreateApply,
     TableImport,
+    ExternalTableCreate,
 }
 
 impl std::fmt::Display for JobKind {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            JobKind::Unspecified => write!(f, "Unknown"),
+            JobKind::Unspecified => write!(f, "Unspecified"),
```
colinmarc: I actually think `Unknown` makes more sense to end-users, but up to you.

colinmarc (approved):

PR #404, opened 2026-08-21
Avoid deletion of checked-out branch

### LINKED ISSUE #400: Bauplan probably shouldn't let you delete the branch you're currently on
**What happened**

```
bauplan checkout -b user.foo
bauplan branch rm user.foo        # deletes active branch, no warning
bauplan branch create user.bar    # fails
Error: Unknown error (404): Ref "user.foo" not found (REF_NOT_FOUND)
```

After deleting the branch I was on, every `branch create` / `checkout -b` fails with REF_NOT_FOUND for the *old* branch, regardless of the new name. Confusing because the error mentions a branch I'm not referencing.

Workaround: `bauplan checkout main` first, then everything works.

**Suggested fix**

Git solves this at delete-time: `git branch -d` refuses to delete the checked-out branch, even with force. Recommend `branch rm` does the same (or auto-checkout main with a warning).

Failing that, commands that hit a dangling active branch should detect it and say "your active branch no longer exists, run `bauplan checkout main`" instead of a 404 about a ref I never typed.

**Bonus**: cosmetic warning alongside it — `WARN Failed to parse API error kind: REF_NOT_FOUND.input_ref: invalid type: map, expected a string`

CLI: bauplan 0.1.16 (pip), macOS

giacomopiccinini:
This PR fixes the current behavior which lets users delete the Bauplan branch they are on (see [this](https://github.com/BauplanLabs/bauplan/issues/400)) without raising an error/warning, and causing downstream problem when trying to create a new branch.  

Closes #400.

### DIFF
--- a/src/cli/branch.rs
+++ b/src/cli/branch.rs
@@ -333,9 +333,7 @@ fn create_branch(cli: &Cli, args: BranchCreateArgs) -> anyhow::Result<()> {
     }
 
     eprintln!("Created branch \"{branch_name}\"");
-    anstream::eprintln!(
-        "{GREEN}TIP:{GREEN:#} To switch to the new branch, run:",
-    );
+    anstream::eprintln!("{GREEN}TIP:{GREEN:#} To switch to the new branch, run:");
     eprintln!("\tbauplan checkout {branch_name:?}");
     Ok(())
 }
@@ -347,6 +345,10 @@ fn delete_branch(cli: &Cli, args: BranchRmArgs) -> anyhow::Result<()> {
     } = args;
 
     let req = DeleteBranch { name: &branch_name };
+
+    if cli.profile.active_branch.as_deref().unwrap_or("main") == branch_name.as_str() {
+        bail!("cannot delete the active branch {branch_name:?}");
+    }
 
     if let Err(e) = cli.roundtrip(req) {
         if if_exists && matches!(api_err_kind(&e), Some(ApiErrorKind::BranchNotFound { .. })) {
--- a/tests/cli/branch.rs
+++ b/tests/cli/branch.rs
@@ -1,3 +1,4 @@
+use crate::cli::config::config_set;
 use crate::cli::{bauplan, test_branch, username};
 use predicates::prelude::PredicateBooleanExt as _;
 use predicates::str::{contains, starts_with};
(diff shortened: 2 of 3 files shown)


src/cli/branch.rs
```diff
@@ -348,6 +346,10 @@ fn delete_branch(cli: &Cli, args: BranchRmArgs) -> anyhow::Result<()> {
 
     let req = DeleteBranch { name: &branch_name };
 
+    if cli.profile.active_branch.as_deref().unwrap_or("main") == branch_name.as_str() {
+        bail!("Cannot delete the branch \"{branch_name}\" which you are currently on.");
```
bigluck: we usually say `active branch` elsewhere, and the branch errors nearby use `{:?}` and no trailing period; what about something like:

```
cannot delete the active branch {branch_name:?}
```

bigluck: maybe you could also add a tip like the one above, so the user knows the way out:

```
TIP: switch to another branch first, e.g.:
\tbauplan checkout main
```

src/cli/branch.rs
```diff
@@ -333,9 +333,7 @@ fn create_branch(cli: &Cli, args: BranchCreateArgs) -> anyhow::Result<()> {
     }
 
     eprintln!("Created branch \"{branch_name}\"");
-    anstream::eprintln!(
-        "{GREEN}TIP:{GREEN:#} To switch to the new branch, run:",
-    );
+    anstream::eprintln!("{GREEN}TIP:{GREEN:#} To switch to the new branch, run:",);
```
bigluck: nit: the dangling comma can be dropped now that this is a single line

src/cli/branch.rs
giacomopiccinini: Sure, I can do that. I was just replicating the GH message.

giacomopiccinini: @colinmarc @bigluck now it prints

```
Error: cannot delete the active branch "<branch>"
TIP: switch to another branch first, e.g.:
	bauplan checkout main
```

What do you think?

src/cli/branch.rs
```diff
@@ -348,6 +346,14 @@ fn delete_branch(cli: &Cli, args: BranchRmArgs) -> anyhow::Result<()> {
 
     let req = DeleteBranch { name: &branch_name };
 
+    if cli.profile.active_branch.as_deref().unwrap_or("main") == branch_name.as_str() {
+        bail!(
+            "cannot delete the active branch {branch_name:?}\n\
+                {GREEN}TIP:{GREEN:#} switch to another branch first, e.g.:\n\
```
colinmarc: Does this work with `anstream`? I wouldn't expect it to.

src/cli/branch.rs
giacomopiccinini: Moving back to `Error: cannot delete the active branch "<branch>"`

colinmarc (approved):

bigluck (approved):

PR #405, opened 2026-08-21
docs: split ci jobs

innazh:
## Problem
When PRs come from a forked repo - the credentials needed to successfully run CI are missing, so the CI is fails.

## Solution
Splitting up the CI tasks more granularity, so that at least we can run a build tasks on the changes, without the deploy one, and lint without the integrations one. We can run tasks that don't require the credentials and skip the tasks that require them if it's a forked PR.

## Changes:
- tests.yaml: split the test job into lint and test-integration, the latter
skipped on fork PRs.

- docs.yaml: split the deploy job into build, which uploads the site as an
artifact, and deploy, which downloads it and does the AWS work. deploy
holds id-token: write and skips on fork PRs.

- Justfile: add a test-integration recipe.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,10 +13,9 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test:
-    name: Test
+  lint:
+    name: Lint
     runs-on: blacksmith-16vcpu-ubuntu-2404
-    environment: test
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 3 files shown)


colinmarc:
In the README we say that we're not really looking for outside contributions. I personally wouldn't change anything to enable them.

innazh:
> In the README we say that we're not really looking for outside contributions. I personally wouldn't change anything to enable them.

Valid point, I didn't know that would be the case either until apo asked be to review an PR with outside contributions. This may have been just a one off event though.

.github/workflows/tests.yaml
```diff
@@ -29,7 +28,23 @@ jobs:
 
       - uses: ./.github/actions/setup-caches
 
-      - run: mise x -- just test
+      - run: mise x -- just lint
+
+  test-integration:
+    name: Integration Tests
```
drin, line 34:
```suggestion
    name: Integration Tests
    needs: lint
```

Should this job depend on the `lint` job?

drin (commented):
left a comment about `test-integration` maybe depending on `lint` (if I am reading it correctly it does not yet).

everything else looks straightforward to me, but I'll let @bigluck provide actual blessing

drin:
> I personally wouldn't change anything to enable them.

I wouldn't go out of our way to enable them, but this PR doesn't seem so bad. If nothing else we can add the appropriate check to every CI job to skip on fork PRs?

bigluck (changes requested):
Ciao @innazh, thanks so much for this PR and the effort you've put into it, but I'd rather close it and not add fork support; after some internal discussion, the CI needs credentials that forks can't get, and I don't want to risk silently skipping tests on our own PRs. The `README.md` in the root already says we're not looking for outside contributors, but GitHub issues are the right way to report bugs.

Anyway, if we want Omer's changes, we can just cherry-pick them onto an internal branch with something like:

```bash
git fetch https://github.com/omerbensaadon/bauplan.git patch-1
git checkout main && git pull
git checkout -b omer/patch-1
git cherry-pick FETCH_HEAD
git push -u origin omer/patch-1
```

...this keeps Omer as the author and from there you can open a new PR and CI runs as expected, without additional changes, but just keep in mind that by doing this external code will run in our CI with our credentials, so review the diff carefully before pushing, because the responsibility for what gets executed is on us, not on the contributor.

CC @colinmarc & @jacopotagliabue

PR #406, opened 2026-08-25
Improve compile times

colinmarc:
This removes a bunch of polyglot-sql features, and bumps other dependencies.


jacopotagliabue (approved):
sembra un timbrone!

PR #423, opened 2026-09-02
fix: alpha stub dependency

drin:
Depend on sdk stub starting with alpha version 0.1.0a1


drin:
closing in favor of #424

PR #425, opened 2026-09-02
fix: temporarily simplify test

drin:
This temporarily fixes a test that isn't supported by the runtime

### DIFF
--- a/tests/fixtures/multi-field-schema/models.py
+++ b/tests/fixtures/multi-field-schema/models.py
@@ -83,7 +83,13 @@ def multi_field_model(
     ],
     locations: Annotated[
         pyarrow.Table,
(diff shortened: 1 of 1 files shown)


tests/fixtures/multi-field-schema/models.py
```diff
@@ -83,7 +83,13 @@ def multi_field_model(
     ],
     locations: Annotated[
         pyarrow.Table,
-        Model("query_model", projection_schema=LocationsPushdown),
+        Model(
```
bigluck, line 86: can we fix it upstream?

bigluck: or downstream.. I mean in our runner :)

drin: absolutely

PR #426, opened 2026-09-04
feat: extend CLI and pySDK to retrieve table and column documentation

drin:
Extends the `table get` CLI command to also show the table and column documentation. Adds attributes to make table and column docs available to pySDK.

For long column documentation or documentation beyond a newline, the CLI truncates the remaining documentation and prints `...`.


src/api/iceberg.rs
```diff
@@ -11,6 +11,31 @@ use crate::{
...
+impl ApiRequest for LoadTable<'_> {
+    type Response = IcebergTable;
+
+    fn path(&self) -> PathArgs {
+        urlformat!(
+            "/iceberg/v1/{}/namespaces/{}/tables/{}",
+            self.at_ref,
+            self.namespace,
+            self.name,
+        )
```
bigluck: I'm not sure this is the right approach; why don't we extend the response produced by our API server?

drin: this was mostly a proof-of-concept. I will have the catalog changes in a PR momentarily and then I'll update this to accommodate it

colinmarc:
Can we just add this information in a reasonable way to `table get` instead of adding a new command?

drin:
i must've just forgotten to remove the LoadTable stuff, docs should be coming back in catalog response for get table already

tests/fixtures/invalid_package_pppandas/models.py
```diff
@@ -16,32 +16,39 @@
...
-    dropoff_datetime: TimestampMicroUTC
-    PULocationID: Int64
-    DOLocationID: Int64
-    trip_miles: Float64
-    trip_time: Int64
-    base_passenger_fare: Float64
-    tolls: Float64
-    sales_tax: Float64
-    tips: Float64
+    pickup_datetime: TimestampMicroUTC | None
```
colinmarc: Why did all these fields become nullable? Why is this in a PR about `table_get`?

drin: removed from this PR

src/cli/table.rs
```diff
@@ -428,17 +428,26 @@ fn handle_get_table(
...
                 r#type,
+                doc,
                 ..
             } in resp.fields
             {
-                writeln!(&mut tw, "{name}\t{required}\t{type}")?;
+                writeln!(
+                    &mut tw,
+                    "{name}\t{required}\t{type}\t{}",
+                    doc.as_deref().unwrap_or("-")
```
bigluck, line 449: uhm, what happens if the user has a multi-line docstring?

drin: alright, I tried a few options, and updated the PR description to mention them. I'll resolve this thread and then we can open new comments as needed

src/cli/table.rs
```diff
@@ -408,6 +408,34 @@ fn handle_list_tables(
     Ok(())
 }
 
+/// Condense documentation into a single cell, cut at the first newline or at
+/// `DOC_WIDTH` characters, whichever comes first.
+///
+/// Anything dropped is marked with `...`; the full text is available from
+/// `--output json`.
+fn doc_summary(doc: Option<&str>) -> String {
```
vtk9: i think this function could use a unit test - so the reader can see how it gets truncated. esp if the comment doc summary is multiline

src/cli/table.rs
```diff
@@ -408,6 +408,34 @@ fn handle_list_tables(
...
+    // Tabs are expanded because the tab writer reads them as cell separators.
+    let mut lines = doc.lines();
+    let first = lines.next().unwrap_or_default().replace('\t', " ");
+    let first = first.trim_end();
+
+    // The cell holds the whole doc only if nothing follows the first line and
+    // that line fits.
+    if lines.next().is_none() && first.chars().count() <= DOC_WIDTH {
+        return first.to_string();
+    }
```
bigluck, line 432: nit: what about something like:

```rust
let (first, rest) = doc.split_once('\n').unwrap_or((doc, ""));
let first = first.trim_end().replace('\t', " ");

if rest.is_empty() && first.chars().count() <= DOC_WIDTH {
    return first;
}
```

src/cli/table.rs
bigluck: nit: `print_tty` for the query in `query.rs` has the code for truncating the string inside the function, with `TRUNCATE_TO_COLUMN_WIDTH = 32` and a note on stderr when something gets cut ("Note: some values were truncated..."). Maybe we could be consistent with that we already have for the query.

https://github.com/BauplanLabs/bauplan/blob/main/src/cli/query.rs#L259-L355

drin: sure, I can do that. do we want to allow `-no-trunc`?

drin: done

bigluck: Let's start without, we could always add it in the future... json output already prints out the full doc

src/cli/table.rs
drin: should be done

src/cli/table.rs
drin: I use the same width, and I changed `doc_summary` to return a `(bool, String)` so that the `bool` can be used to print the same "something was truncated" message that `print_tty` does.

I noticed that `print_tty` also prints the schema header as `"COLUMN\tTYPE\tNULLABLE"` for `query` compared to `"NAME\tREQUIRED\tTYPE"` for `table get`. Do we want to make those the same?

drin: I made `query` and `table get` schema output consistent in terms of column headers:
- COLUMN
- TYPE
- REQUIRED

I did not add documentation to the schema output of the `query` command because we do not yet support SQL to modify the documentation so it can be confusing to see it for most queries that aren't just straight reads.

drin: I replaced the naive unit tests I had with an integration test that covers the various cases

tests/cli/table.rs
```diff
@@ -81,6 +82,40 @@ fn get_json_output() {
         .stdout(contains(r#""namespace":"bauplan","#));
 }
 
+#[test]
+fn get_column_documentation() {
```
bigluck, line 86: I like the integration test, but I would not remove the unit tests because they test different things. `doc_summary` is a pure fn and the old unit tests checked the tab replace, the cut on multibyte chars and the whitespace before ...; this integration test does not check any of these and in some cases can't check them.

For example the tab in `Fare\tbefore` should never reach the cli when the doc comes from a bauplan model, but a user can update the table definition using our iceberg endpoints, so I think it's important to validate that case too, in the unit tests.

src/cli/query.rs
```diff
@@ -269,14 +269,14 @@ async fn print_tty(
     {
         let mut tw = TabWriter::new(&mut stdout);
 
-        writeln!(tw, "COLUMN\tTYPE\tNULLABLE")?;
+        writeln!(tw, "COLUMN\tTYPE\tREQUIRED")?;
```
bigluck, line 272: I'm skeptical about required here :( because if I look at our stack, our base system is always `arrow`; our type contract types mirror the arrow types (`Int64`, `Float64`, `Date32`, `TimestampMicro` etc...); the user writes `Int64 | None` which means `nullable`, and arrow has `nullable`, not required.
Same for Delta (`nullable`) and SQL (`NOT NULL`); the only place where we have required is iceberg, because today the catalog returns the iceberg schema, but tomorrow with the new api (the contract is already defined in bpln-api) will return the columns as arrow fields, with `nullable`.

So with the change here and in the `query.rs` file, the query command reads an arrow schema and prints it with the `iceberg` word, and I find it confusing.

Uniforming the two headers is perfect and I like it, but maybe the direction should be the opposite, in other words I would prefer (personal taste) `NULLABLE` in both.

src/cli/table.rs
```diff
@@ -408,6 +411,41 @@ fn handle_list_tables(
     Ok(())
 }
 
+/// Condense documentation into a single cell, cut at the first newline or at
+/// `TRUNCATE_TO_COLUMN_WIDTH` characters, whichever comes first.
+///
+/// Anything dropped is marked with `...`; the full text is available from
+/// `--output json`.
+fn doc_summary(doc: Option<&str>) -> anyhow::Result<(bool, String)> {
```
bigluck, line 419: nit: since `doc_summary` is a pure fn I would not return an `anyhow::Result` here (what about using `format!` instead of `write!`?)

src/cli/query.rs
bigluck, line 272: @drin can we move the `query.rs` change to its own commit? It's a visible change on another command and the changelog is built from the commit titles; And can you also please update the PR description too? merci

giacomopiccinini: I think I agree with you. Another byproduct is that there's a little bit more cognitive load for the user because `required=True` corresponds to `nullable=False`, and in the past it generated a little headache at first. I think users in general reason around the concept of "nullable" and not of "required"?

jacopotagliabue: ok, go on!

drin: good point, I forgot about the changelog. but yeah o can split the commit easily

src/cli/table.rs
drin: sure, I did write for consistency with print_tty, but i can change it easily.

drin: sure, sounds good

tests/cli/table.rs
drin: the integration test does check the whitespace before `...` and it can check multibyte chars if I update the doc strings.

To cover these cases, claudio changed the doc string for `trip_miles`: changed `The distance travelled while the passenger was in the vehicle.` to `Distanza percorsa: è espressa in miglia.` (this is claudio translated, I can't be blamed for poor wording).

Because that was the case that covered a truncation occuring right after whitespace, I changed the doc string for `dropoff_datetime`: `The time the passenger left the vehicle.\n` changed to `The time the passengers left the vehicle.\n` so that the truncation occurs after `left` (` the vehicle.\n` -> ` ...`; then ` ...` -> `...`).

I re-added the tab expansion unit test, but I don't understand why we don't store tab characters in the catalog

src/cli/table.rs
```diff
@@ -24,6 +24,8 @@ use indicatif::ProgressBar;
 use tabwriter::TabWriter;
 use tracing::info;
 
+const TRUNCATE_TO_COLUMN_WIDTH: usize = 32;
```
vtk9: this constant is used in 2 places with the same value (one in query.rs) we should either create a shared place - or add a comment that anyone modifying `TRUNCATE_TO_COLUMN_WIDTH` here should also remember to modify it in `query.rs` (and a add a comment there)

python/bauplan/__init__.pyi
```diff
@@ -1076,9 +1076,12 @@ class Client:
         # You can get the total number of rows this way.
```
vtk9, line 1076: nit: the commit message is mentioning get table but there's only `bauplan table get`

vtk9 (dismissed):

src/cli/table.rs
```diff
@@ -428,20 +464,34 @@ fn handle_get_table(
             println!();
         }
         Output::Tty => {
+            if let Some(comment) = resp.comment() {
```
bigluck, line 467: nit: `comment()` may return `Some("")` when the property exists but is empty; bauplan removes the property when it is bauplan itself writing the table, but an external query engine can update an external table via the iceberg endpoints, so I think the cli should treat the table comment as a value we don't control; I suggest the same trim + empty check as `doc_summary` maybe inside `Table::comment()` so the python getter returns None too.

src/cli/table.rs
```diff
@@ -428,20 +464,34 @@ fn handle_get_table(
...
+                let (is_doc_truncated, field_doc) = doc_summary(doc.as_deref());
+                is_truncated = is_truncated || is_doc_truncated;
+                writeln!(&mut tw, "{name}\t{type}\t{}\t{field_doc}", !required)?;
             }
 
             tw.flush()?;
+
+            if is_truncated {
+                eprintln!(
+                    "\nNote: some documentation was truncated. Use json output to see full documentation."
```
bigluck: nit: I would name the flag here like query does and make it shorter; what about something like:

> Note: some documentation ws truncated, use -O json to see it in full.

bigluck (approved):

src/cli/table.rs
drin: I think that's motivation that an empty string shouldn't unset the documentation in bauplan, that way we don't have different behavior of setting an empty string in bauplan vs setting an empty string via iceberg

src/cli/table.rs
drin: I use it for consistency, but I'm not sure we need to actually make them the same, because at the moment query does not return documentation but `table get` does. I'll likely lean towards a comment, but I'll see what makes sense

src/cli/table.rs
drin: okay, I created a free function, `clean_doc` that gets called in both `Table::comment()` and `doc_summary`. It trims the given &str and if whitespace-only it downgrades to `None`.

src/cli/table.rs
drin: I did end up thinking a comment was best, but I reference cli::query::TRUNCATE_TO_COLUMN_WIDTH and say that the same value is used for consistency but since `table get` is the only command showing documentation we may arbitrarily want to widen it.

drin: I also renamed the variable so that it's not semantically tied to "COLUMN_WIDTH" considering if we do widen that particular width it will be wider than the other columns

python/bauplan/__init__.pyi
drin: changed the commit message

vtk9 (approved):

PR #427, opened 2026-09-05
fix: add `ty` ignore for `bauplan init`

drin:
Added ty ignore directive for import in init DAG.

### DIFF
--- a/src/cli/init/models.py
+++ b/src/cli/init/models.py
@@ -69,7 +69,7 @@ def survival_rate_by_age(
     | 20       | 0.4           |
     """
 
-    import polars as pl
+    import polars as pl  # ty: ignore[unresolved-import]
 
     df = pl.DataFrame(passengers)
     return (


giacomopiccinini:
@colinmarc @bigluck I asked @drin to bring back the comment preventing ty from raising an error on the unresolved polars import. We used to have `# noqa` (which should be meant for ruff, and not ty, I think), but the ty-compatible way is `# ty: ignore[unresolved-import]` (if we want to be ty-specifc/compliant). An alternative is `# type: ignore` which is not ty-specific and should work with other checkers, too.

What do you guys think?

colinmarc:
It would be easier to just fix it by adding polars to the generated `pyproject.toml`, no?

giacomopiccinini:
> It would be easier to just fix it by adding polars to the generated `pyproject.toml`, no?

That would solve it, yes, but I feel like it's not really what we expect users to do in general? For instance, I could do `uv add polars==1.44`, but then have `bauplan.python(pip={"polars": "1.42"})` in the `models.py`. My point being that the two might not be in sync, and I think we do not ask users to add packages to the bauplan project's `pyproject.toml`

colinmarc:
> > It would be easier to just fix it by adding polars to the generated `pyproject.toml`, no?
> 
> That would solve it, yes, but I feel like it's not really what we expect users to do in general? For instance, I could do `uv add polars==1.44`, but then have `bauplan.python(pip={"polars": "1.42"})` in the `models.py`. My point being that the two might not be in sync, and I think we do not ask users to add packages to the bauplan project's `pyproject.toml`

I do want to see if we can rely on uv and lockfiles for model dependencies - but you're right that that's not the case today.

giacomopiccinini (approved):

colinmarc (approved):

PR #429, opened 2026-09-08
feat: tweaks to the @python decorator and fixtures

colinmarc:
These are some miscellany I found while working on the new parser:

 - `@python()` without a version doesn't make much sense.
 - Quoting a whole annotation as a forward expr is redundant, since we treat annotations as lazy (as does python by default from 3.14)

### DIFF
--- a/bauplan-sdk-types/src/bauplan_sdk_types/_runtimes.py
+++ b/bauplan-sdk-types/src/bauplan_sdk_types/_runtimes.py
@@ -4,7 +4,7 @@ from typing import Optional, Callable
 
 
 def python(
-    version: Optional[str] = None,
+    version: str,
     pip: Optional[dict[str, str]] = None,
 ) -> Callable:
     """
--- a/tests/fixtures/missing-schema/models.py
+++ b/tests/fixtures/missing-schema/models.py
@@ -26,5 +26,5 @@ def typed_params_model(
             filter="PULocationID = 138",
         ),
     ],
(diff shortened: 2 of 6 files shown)


drin:
> we treat annotations as lazy (as does python by default from 3.14)

The only thing that matters is the python runtime (because we do treat them "lazily"). as long as we support python >=3.11, we should be able to accept annotations as forward expressions.

bauplan-sdk-types/src/bauplan_sdk_types/_runtimes.py
```diff
@@ -4,7 +4,7 @@
 
 
 def python(
-    version: Optional[str] = None,
+    version: str,
```
bigluck: isn't it a `feat!` (space) ?

colinmarc: Yeah, you're right.

bigluck:
@colinmarc `tests/fixtures/syntax-smoke/models.py:197` still has `@python()`

bigluck:
It looks good to me, but we can not merge it now. Can you make it a draft as I did for:
- #462

colinmarc:
Instead, I took out the breaking change - I'll make another PR with all the breaking changes.

bigluck (approved):

colinmarc:
Follow-up: https://github.com/BauplanLabs/bauplan/pull/475

PR #431, opened 2026-09-09
feat(pysdk): drop the title parameter of TableField

bigluck:
Iceberg has a single `doc` per column, so we keep a single `doc` on `TableField` to be consistent with the underlying system where the string gets stored.

### DIFF
--- a/bauplan-sdk-types/pyproject.toml
+++ b/bauplan-sdk-types/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "bauplan-sdk-types"
-version = "0.1.0"
+version = "0.2.0a1"
 requires-python = ">=3.11"
 readme = "README.md"
 description = "Type and decorator definitions for authoring Bauplan pipelines"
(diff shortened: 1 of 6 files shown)


bauplan-sdk-types/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "bauplan-sdk-types"
-version = "0.1.0"
+version = "0.2.0a1"
```
colinmarc, line 7: I think we have to change the pin in the root pyproject.toml?

colinmarc (approved):

bauplan-sdk-types/pyproject.toml
bigluck: shouldn't we merge the PR first, release to pypi and then open a PR for the pin?

colinmarc: I don't really think we need to do that separately, no. The pin doesn't matter until we publish `bauplan`.

PR #432, opened 2026-09-09
Update examples to sdk 0.3.0

giacomopiccinini:
Update all the exisiting examples to v0.3.0.

PR #445, opened 2026-09-09
fix: Update schema specifications for resources

drin:
A variety of fixes to accommodate the release of the new type contracts feature that allows users to define schema specifications.

Fix checklist:
- [x] In sample project (`bauplan init`), made a column optional to satisfy schema specifications.


vtk9 (approved):

drin:
the failing test is addressed in #432 but this passes when I run init and then run the project itself and ruff and ty check it

PR #450, opened 2026-09-10
docs: update expectations docs

giacomopiccinini:
Specify that expectations are enforced by default.

### DIFF
--- a/docs/pages/concepts/expectations.mdx
+++ b/docs/pages/concepts/expectations.mdx
@@ -224,69 +224,11 @@ def test_data_freshness(
     return check_data_recency(data, 'timestamp', hours=24)
 ```
 
-## Failure handling strategies
+## Handling expectation failures
 
-Expectations offer flexible failure handling to match the severity of different data quality issues.
+By default, Bauplan enforces expectations: if any expectation fails, the pipeline halts immediately.
 
-### Hard failures (assert)
-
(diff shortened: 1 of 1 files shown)


docs/pages/concepts/expectations.mdx
```diff
@@ -224,69 +224,11 @@ def test_data_freshness(
...
-def test_data_freshness(
-    data: Annotated[pyarrow.Table, bauplan.Model('model', projection_schema=Timestamps)],
-) -> bool:
-    is_fresh = check_data_recency(data, 'timestamp', hours=24)
-
-    return is_fresh
-```
-
-**Use for**: issues you want to monitor but that don't require immediate pipeline termination, allowing you to collect metrics and set up alerts.
+When failures are not critical, you can log them and let the pipeline continue by disabling strict mode on the `run` command: pass `--no-strict` in the [CLI](https://docs.bauplanlabs.com/reference/cli#bauplan-run) or `strict=False` in the [Python SDK](https://docs.bauplanlabs.com/reference/bauplan#bauplan-client-run).
```
bigluck: why here we have absolute urls? Our docs get deployed in `dev` too, and with these links we can not test them there

giacomopiccinini: Because I am stupid. Should be fixed now.

bigluck (approved):
merci

PR #452, opened 2026-09-10
docs: fix syntax for partitioned_by in SQL

giacomopiccinini:
Fix partitioned_by syntax with underscore, and remove quotes. 

Tested on 0.3.0. Originally encountered by Timo Dechau.


bigluck (approved):
nit: can you test if it works with the space between the comma and the 2nd column?

giacomopiccinini:
> nit: can you test if it works with the space between the comma and the 2nd column?

It does @bigluck

PR #434, opened 2026-09-09
fix(sdk): replace fields with on/off with bool, and make strict mode default

giacomopiccinini:
Make the SDK adhere to the CLI convention where strict is bool flag

PR #455, opened 2026-09-10
feat(pysdk)!: use a boolean cache option on query and scan

bigluck:
`Client.run` moved to `cache: bool` in 0.3.0-rc.2, but `query`, `scan` and `query_to_*` still wanted `cache='on'|'off'`, so `client.query(cache=False)` failed with a `ValueError`, and this inconsistency is not great from the user pov.

Now every method takes the same `bool`, defaulting to `True` like the CLI `--no-cache` flag.

Follow up of:
- #434

### DIFF
--- a/examples/learn/05-advanced-git-for-data/git_for_data_tour.py
+++ b/examples/learn/05-advanced-git-for-data/git_for_data_tour.py
@@ -83,7 +83,7 @@ def main(file_path: str):
     run_1 = client.run(
         project_dir="./my_project",
         ref=my_branch_name,
-        cache="off",
+        cache=False,
         parameters={"run_id": 1},
     )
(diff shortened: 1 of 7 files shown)


bigluck:
@giacomopiccinini & @colinmarc I know very well that it is a breaking change, but I vote to release it in the next v0.3.x

colinmarc:
Is this actually urgent in any way?

giacomopiccinini (approved):
Approved, and agreed on having it as 0.3.x.

bigluck:
@colinmarc I'm packaging up some more similar changes. I'd like to release them once I've compiled everything... I have a pretty extensive report to review, for now I'm good if we merge it without immediately releasing it to pypi

PR #457, opened 2026-09-10
test: typecheck the docstring examples of bauplan-sdk-types

bigluck:
The `docstrings` test only scanned `python/bauplan/*.pyi`, so the examples in
`bauplan-sdk-types` reached the reference docs unchecked; the `expectation()` one
had a trailing comma that made it a SyntaxError.

### DIFF
--- a/bauplan-sdk-types/src/bauplan_sdk_types/_parameters.py
+++ b/bauplan-sdk-types/src/bauplan_sdk_types/_parameters.py
@@ -7,11 +7,11 @@ from dataclasses import dataclass
 class Parameter:
     """
     Represents a parameter that can be used to "template" values passed to a model during a run or
-    query with, e.g., ``bauplan run --parameter interest_rate=2.0``.
+    query with, e.g., `bauplan run --param interest_rate=2.0`.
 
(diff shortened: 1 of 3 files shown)


bauplan-sdk-types/src/bauplan_sdk_types/_parameters.py
```diff
@@ -7,7 +7,7 @@
 class Parameter:
     """
     Represents a parameter that can be used to "template" values passed to a model during a run or
-    query with, e.g., ``bauplan run --parameter interest_rate=2.0``.
+    query with, e.g., ``bauplan run --param interest_rate=2.0``.
```
colinmarc: Why double backticks?

colinmarc (approved):
Good catch, thanks!

bauplan-sdk-types/src/bauplan_sdk_types/_parameters.py
bigluck: isn't it the RST style inherited from the old pysdk docstring?

bigluck: but I think we have them only here... fixing, thanks

PR #458, opened 2026-09-10
Inna/docs 0.3.0 pass

innazh:
- Updating the code in docs/pages/common-scenarios/multi-stage-pipelines.mdx.
- Adding tabs with sql and duckDB examples side by side to the models page.
- Removing a redundant example

### DIFF
--- a/docs/pages/concepts/models.mdx
+++ b/docs/pages/concepts/models.mdx
@@ -22,7 +22,11 @@ A typical Python model consists of two decorators `@bauplan.model()` and
 `@bauplan.python()` and a function.
 
 ```python
-#! from typing import Annotated
+from typing import Annotated
+
+import bauplan
+import pyarrow
+
 class SelectedInputColumns(bauplan.TableSchema):
     """Columns read from the input table."""
 
@@ -36,40 +40,46 @@ def my_model(
         data: Annotated[
             pyarrow.Table,
             bauplan.Model(
-                'input_table',
+                "input_table",
                 projection_schema=SelectedInputColumns,
                 filter="timestamp >= '2022-12-15T00:00:00-05:00'",
             )
         ],
 ) -> Annotated[pyarrow.Table, SelectedInputColumns]:
-    ...
-
-    #! output_table = pyarrow.table({})
-    return output_table
+    return data
 ```
 
 Bauplan Models are **fully declarative**, allowing for explicit column
 selection and filter pushdown. In practice, this means your code only
 requires the name of the inputs and, optionally, the desired columns and
-filters. Naming the columns and filters you need lets the planner skip
-reading the rest of the table.
+filters.
 
-For example, you don't need to specify the type of `input_table`,
-whether it's an Iceberg Table, PyArrow Table, Pandas dataframe, or
-Polars dataframe. This approach makes the code fully portable, easier to
-reproduce across environments, and simpler to maintain.
+For example, you don't need to specify whether `input_table` is an Iceberg Table, PyArrow
+Table, Pandas dataframe, or Polars dataframe. This approach makes the code fully
+portable, easier to reproduce across environments, and simpler to maintain.
 
-:::note Filters only apply to catalog inputs
+:::note Filters only apply to models from the lakehouse catalog
 
-`filter` only works on an input that reads from the catalog. Filtering an input
-that is another model in the same pipeline is not valid syntax, and the planner
-rejects it.
+`filter` only works on an input that reads from the catalog. Filtering an input that is
+another model in the same pipeline is not supported, and the planner rejects it.
 
-For example, `bauplan.Model('orders', filter='amount > 100')` is valid when
-`orders` is a catalog table, but the same filter on
-`bauplan.Model('purchases', ...)` is rejected when `purchases` is a model in the
-same pipeline. To narrow the rows `purchases` passes along, apply the filter
-inside `purchases` and return the filtered table.
+For example, `my_model` applies a filter on `timestamp` on
+`bauplan.Model('input_table')`. This is valid because `input_table` is not defined in the
+pipeline and must come from the catalog. It is not supported for a new model,
+`other_model`, to take `my_model` as an input and apply a filter to it.
+
+```python type:ignore
+@bauplan.model()
+@bauplan.python('3.11')
+def other_model(
+    data: Annotated[
+        pyarrow.Table,
+        # filters can only be applied to models from the catalog
+        bauplan.Model("my_model", filter="timestamp <= '2022-12-16'")
+    ]
+) -> Annotated[pyarrow.Table, SelectedInputColumns]:
+    return data
+```
 
 :::
 
@@ -126,11 +136,13 @@ you declare for that model alone. Declare only what the function body
 imports.
 
 ```python
-#! from typing import Annotated
+from typing import Annotated
+import bauplan
+
 class StepOutput(bauplan.TableSchema):
     """The columns each step passes along."""
 
-    ...
+    col: bauplan.Int64
 
 @bauplan.model()
 @bauplan.python('3.11', pip={'pandas': '2.2.0'})
@@ -138,9 +150,12 @@ def step_1(
     data: Annotated[pyarrow.Table, bauplan.Model('input_table')],
 ) -> Annotated[pyarrow.Table, StepOutput]:
     import pandas
+
+    df = data.to_pandas()
+
     ...
 
-    #! return pyarrow.table({})
+    return pyarrow.table(df[['col']])
 
 @bauplan.model()
 @bauplan.python('3.11', pip={'pandas': '1.5.3'})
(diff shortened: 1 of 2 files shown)


docs/pages/concepts/models.mdx
```diff
@@ -245,64 +290,15 @@ def my_model(
...
-   - Separate function bodies into external modules and call them within the Bauplan models. This keeps business logic code neatly separated from the DAG and environment declaration, making code refactoring easier and future-proof.
-
-5. **Testing and Development**
-   - Use the `bauplan run --dry-run` flag for quick iteration.
-   - Employ print statements for real-time terminal feedback during development.
+  </TabItem>
+</Tabs>
+
+See [SQL files](/common-scenarios/sql-files) for the other header options
+and for including `.sql` files that live outside the project root.
```
giacomopiccinini: possibly misleading, because we only look "downward" from the project root.

docs/pages/concepts/models.mdx
```diff
@@ -202,11 +210,48 @@ Requirements:
...
+
+```sql title="my_model.sql"
+-- bauplan: output_schema=CountByColOne
+SELECT col_1, COUNT(*) AS count
+FROM input_table
+WHERE timestamp >= '2022-12-15T00:00:00-05:00'
+GROUP BY col_1
+```
+
+Bauplan discovers `.sql` files in the project root and takes the model name
```
giacomopiccinini: I think we can slightly rephrase this. The point is: it discovers `.sql` in the project root _automatically_, but you can possibly include more nested paths with the `include_paths` field in the `bauplan_project.yaml`.

docs/pages/concepts/models.mdx
drin:
```suggestion
and for including `.sql` files in subdirectories below the project root.
```

docs/pages/concepts/models.mdx
```diff
@@ -245,64 +290,15 @@ def my_model(
...
+
+See [SQL files](/common-scenarios/sql-files) for the other header options
+and for including `.sql` files that live outside the project root.
+
+## Developing models
+
+Print statements in a model body stream back to your terminal while the
+run is in flight, which is the quickest way to see what a function
+actually received. Pair them with `bauplan run --dry-run` to iterate
+without materializing anything.
```
drin:
```suggestion
without materializing data to the data lake.
```

I think this small nuance makes the mental model clearer (and I'm hoping it's correct lol)

docs/pages/concepts/models.mdx
```diff
@@ -202,11 +210,48 @@ Requirements:
...
-practice today** for most use cases is to embed your SQL in a Python
-function using DuckDB.
+You can write the model either way. A standalone `.sql` file keeps a pure
+SQL model short. A Python function running DuckDB lets you wrap Python
+around the query, which is what you need to declare pip dependencies or
+to pre- and post-process the data.
 
-```python
+Both declare the same output schema, so a downstream model referencing
+this one cannot tell which you picked.
```
drin:
```suggestion
A SQL model is written as a single SQL query in its own `.sql` file and will be executed by Bauplan's SQL engine. As a Python model, a SQL query is executed in a Python function using an embedded SQL engine of your choice, such as DuckDB or DataFusion. 

Bauplan treats the output of either type of model the same way, for example as an Arrow Table, so a downstream model is agnostic to which language is used to write the model. While a Python model allows you to use SQL directly in the function body, a SQL model can be declared as an input to a separate Python model.
```

This is more in the spirit of what I think we want to say, but maybe you can make it read easier?

docs/pages/concepts/models.mdx
```diff
@@ -202,11 +210,48 @@ Requirements:
...
+class CountByColOne(bauplan.TableSchema):
+    """Row counts grouped by `col_1`."""
+
+    col_1: bauplan.String
+    count: bauplan.Int64
+```
+
+The query lives in a `.sql` file. The optional `output_schema` header ties
+the query to that schema, the way the return annotation does in the Python
+version. Leave it out and the model still runs, but nothing checks its
```
drin:
```suggestion
the query to that schema, the way the return annotation does in a Python
model. Leave it out and the model still runs, but nothing checks its
```

docs/pages/concepts/models.mdx
```diff
@@ -202,11 +210,48 @@ Requirements:
...
+```sql title="my_model.sql"
+-- bauplan: output_schema=CountByColOne
+SELECT col_1, COUNT(*) AS count
+FROM input_table
+WHERE timestamp >= '2022-12-15T00:00:00-05:00'
+GROUP BY col_1
+```
+
+Bauplan discovers `.sql` files in the project root and takes the model name
+from the filename, so this one becomes `my_model`. Downstream models read it
```
drin:
```suggestion
from the filename, so this query is a model named `my_model`. Downstream models read it
```

drin (commented):
a couple changes and then I think this is good!

drin (approved):

docs/pages/concepts/models.mdx
giacomopiccinini, line 57: I am confused by this i think?

giacomopiccinini: @drin

giacomopiccinini, line 72: Why switching from "orders" to "purchases" since we haven't mentioned them before?

giacomopiccinini, line 175: I feel like the Advanced badge is not great, because these days we typically do overwrite partitions all the time

giacomopiccinini, line 179: which tutorial though?

giacomopiccinini, line 212: I think it would be great to also add a comment on the fact that it accepts parameters (which is i think the main use case) and that you can use some syntax on those, i think `day` etc.

docs/pages/concepts/models.mdx
```diff
@@ -245,64 +295,15 @@ def my_model(
...
-   - Separate function bodies into external modules and call them within the Bauplan models. This keeps business logic code neatly separated from the DAG and environment declaration, making code refactoring easier and future-proof.
-
-5. **Testing and Development**
-   - Use the `bauplan run --dry-run` flag for quick iteration.
-   - Employ print statements for real-time terminal feedback during development.
+  </TabItem>
+</Tabs>
+
+See [SQL files](/common-scenarios/sql-files) for the other header options
+and for including `.sql` files in subdirectories below the project root.
```
giacomopiccinini, line 302: this should only appear in the tab of the SQL file i think

docs/pages/concepts/models.mdx
```diff
@@ -202,18 +216,54 @@ Requirements:
...
+class CountByColOne(bauplan.TableSchema):
+    """Row counts grouped by `col_1`."""
+
+    col_1: bauplan.String
+    count: bauplan.Int64
+```
+
+The query lives in a `.sql` file. The optional `output_schema` header ties the query to
+that schema, the way the return annotation does in a Python model. If `output_schema` is
+not specified, the model still runs, but other models cannot use it as an input.
```
bigluck, line 244: @drin hands up on this sentence (more details on slack)

docs/pages/concepts/models.mdx
drin: I think by "type" it means what type of table: iceberg table, pyarrow table, etc. essentially a table is a table (they're all pyarrow tables)

drin: changed to `you don't need to specify whether `input_table` is an iceberg table, ...` etc.

simplified but also this text pre-exists us so I'm not too worried about it right now.

drin: I'll keep it relevant to the example already introduced

drin: removed

drin: no idea. I removed it and changed "will force..." to "will prevent data from being materialized into the lakehouse catalog."

docs/pages/concepts/models.mdx
drin: I originally wrote it, but I fixed it after our discussion and my own investigation. it's actually not an issue, `output_schema` is definitely optional

docs/pages/concepts/models.mdx
drin: done

docs/pages/concepts/models.mdx
drin: you actually pointed out that there are cases where it is likely still an issue. I want to leave the language though and then address in a follow-up if necessary.

docs/pages/concepts/models.mdx
```diff
@@ -202,18 +226,57 @@ Requirements:
...
+class CountByColOne(bauplan.TableSchema):
+    """Row counts grouped by `col_1`."""
+
+    col_1: bauplan.String
+    count: bauplan.Int64
+```
+
+The query lives in a `.sql` file. The optional `output_schema` header ties the query to
+that schema, the way the return annotation does in a Python model. If `output_schema` is
+not specified, then no extra type validation is asserted on the SQL results.
```
bigluck, line 254: I'm now convinced this is a bug... with `projection_schema` on a downstream input a slq model without `output_schema` fails at planning time...this sentence is fine formally, but as discussed I think we should address it upstream first; feel free to double check it

docs/pages/concepts/models.mdx
drin: I will put it in my notion to do list. This may be easy to add but I want the example to be reasonable, etc.

PR #460, opened 2026-09-10
revert(pysdk): use a boolean cache option on query and scan

bigluck:
This reverts commit cde896f7e65fde2e05c1a3159c0ae136cd11315b, merged before we were ready to release 0.4.0.
The change is kept for 0.4.0, so the next release stays on 0.3.x.

The commit is also skipped in `cliff.toml`, otherwise git-cliff keeps computing a 0.4.0 bump from the `feat!` still in the unreleased range.

### DIFF
--- a/.cliffignore
+++ b/.cliffignore
@@ -1,0 +1,2 @@
+# Reverted in 0.3.x, kept for 0.4.0; drop this once v0.3.2 is tagged.
+cde896f7e65fde2e05c1a3159c0ae136cd11315b
--- a/examples/learn/05-advanced-git-for-data/git_for_data_tour.py
+++ b/examples/learn/05-advanced-git-for-data/git_for_data_tour.py
@@ -83,7 +83,7 @@ def main(file_path: str):
     run_1 = client.run(
         project_dir="./my_project",
         ref=my_branch_name,
-        cache=False,
+        cache="off",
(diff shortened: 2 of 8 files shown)


cliff.toml
```diff
@@ -21,6 +21,7 @@ filter_unconventional = true
 tag_pattern = "v[0-9].*"
 include_paths = ["src/**", "python/**", "Cargo.toml"]
 commit_parsers = [
+  { sha = "cde896f7e65fde2e05c1a3159c0ae136cd11315b", skip = true },
```
colinmarc: noooooooooooooooo

colinmarc (commented):
I thought git cliff handles reverts without an explicit blacklist. If it doesn't, then let's just ship 0.4 or break semver.

cliff.toml
bigluck: I know man, It doesn't, see:
- https://github.com/orhun/git-cliff/issues/382 (tested locally too, the revert still bumps to 0.4.0).

The skip is 1 line and goes away after v0.3.2 is tagged. If you prefer, I can move it to a .cliffignore instead of cliff.toml

colinmarc: oh I can deal with .cliffignore. I didn't realize it gets removed, that's also good

colinmarc (approved):

PR #466, opened 2026-09-14
docs: fix spacing on the models page

innazh:
Removing the extra bullet points, and adding a new line before a code block on models page.


bigluck (approved):
👍 thanks!

PR #469, opened 2026-09-16
test: coverage for sql model documentation

drin:
This adds test coverage to `get_documentation` (renamed from `get_column_documentation`) to test:
1. SQL models correctly persist documentation from the `output_schema`
2. A 2nd run of the same project (or, another project that touches the same tables):
   1. clears documentation for columns where `doc` is an empty string or is removed
   2. replaces the table documentation with the schema docstring if the model docstring is removed
   3. if a model is no longer materialized (materialization strategy changes from `REPLACE -> NONE`), that the documentation is not overwritten even if it changed


tests/fixtures/simple_taxi_dag/models.py
```diff
@@ -14,9 +14,23 @@
...
+    + ---------- +   QueryModelSchema    + ----------- +
+    | taxi_fhvhv | --------------------> | query_model |
+    + ---------- +                       + ----------- +
+    """
+
+    pickup_datetime: Annotated[
+        TimestampMicroUTC | None,
+        TableField(
+            doc=(
+                "Pickup time, UTC timezeone, filtered in the range:\n"
```
vtk9: nit: `timezeone` is a typo

vtk9 (approved):

tests/fixtures/simple_taxi_dag_cleardocs/bauplan_project.yml
```diff
@@ -0,0 +1,7 @@
+project:
```
vtk9, line 1: nit: instead of `simple_taxi_dag_cleardocs` maybe name `simple_taxi_incomplete_docs` (up to you)

tests/fixtures/simple_taxi_dag/models.py
drin: ah that explains why it didn't truncate where I expected lol

tests/fixtures/simple_taxi_dag_cleardocs/bauplan_project.yml
drin: I'll keep mostly because the intention is to actually clear documentation that was previously there. incomplete doesn't make that intention as obvious.

PR #471, opened 2026-09-18
feat: Add name property for TableField sdk type

drin:
Update bauplan-sdk-types TableField, adding new `name` property for column names that cannot be used as python identifiers.


bauplan-sdk-types/src/bauplan_sdk_types/_table_fields.py
```diff
@@ -104,10 +104,12 @@ class TableField:
     # TODO: make all parameters keyword args only with `*` arg marker
     def __init__(
         self,
+        name: Optional[str] = None,
```
vtk9, line 137: is this potentially a breaking change if someone is calling TableField("my_doc") without kwargs?

drin: it shouldn't be, because our code analysis doesn't accept positional arguments for `TableField`. The `TODO` is there because adding the `*` arg marker to make everything keyword args only requires a small change to our code analysis (because of how it affects ast parsing).

vtk9: 👍 

how does lineage play a part here? a user might expect to use SampleSchema[`col a`] but it think only `SampleSchema["col_a"]` would work. Should we at least document this behavior?

bauplan-sdk-types/src/bauplan_sdk_types/_table_fields.py
```diff
@@ -99,17 +99,45 @@ class Any(FieldType):
...
+
+    In the above example, `col_a` is an attribute on the class `SampleSchema` and
+    represents a table column. Using the special type, `typing.Annotated`, and this
+    class, `TableField`, it is possible to annotate the represented table column with
+    metadata and constraints.
+
+    The attribute name, `col_a`, corresponds to the name of the represented table column.
+    If the table column's name cannot be used as the attribute name, then the `name`
+    parameter can be specified. For example, if the actual column name is stored with a
+    space: `name='col a''`.
```
vtk9:
```
`name='col a''`.
```
has an extra quote

vtk9 (approved):

bauplan-sdk-types/src/bauplan_sdk_types/_table_fields.py
drin: I put it in the commit message, but I can include it here somewhere

drin: I included it in the `TableField` docstring and I updated the `TableSchema` docstring.

PR #473, opened 2026-09-19
test: Add test coverage for TableField name parameter

drin:
This adds test coverage to `multi-field-schema` test to exercise a column name that needs quoting by using an alias in `query_model.sql`.


tests/fixtures/syntax-smoke/models.py
```diff
@@ -141,7 +143,7 @@ def model_all_options(
         Model(
             "bare_source",
             projection_schema=AnnotatedTypesSchema,
-            filter="annotated_bool = True",
+            filter='"annotated bool" = True',
```
bigluck, line 146: this fixture never reaches the server (it is only imported by `python/tests/test_syntax_stubs.py`), so nothing checks this filter against the backend. Do we have an e2e test somewhere that runs a `filter=` on a column renamed with `name=`? because I couldn't find one, and I think it would be very useful to have, this is the first thing a user with a column like `"fare amount"` will try.

tests/fixtures/multi-field-schema/models.py
```diff
@@ -71,7 +76,10 @@ class LocationsPushdown(bauplan.TableSchema):
 class FullTypesSchema(bauplan.TableSchema):
     """Schema demonstrating all supported field types."""
 
-    pickup_datetime: TimestampMicroUTC | None
+    # Widen this type to be sure validation isn't being short-circuited
```
bigluck, line 79: nit: this repo is public, and I think nobody can understand this comment without knowing how our internals work. Can you describe it from the user's point of view? Something like: "the model returns microseconds and the contract says nanoseconds, so the platform has to convert this column; with the same type the conversion can be skipped".

bigluck (commented):
The fixtures look good to me, I only left 2 small comments... I will approve the PR as soon as the backend PR is merged and deployed to QA and all the tests are green. Thanks!

tests/fixtures/syntax-smoke/models.py
drin: This is just to have coverage. The column isn't renamed, it is already named `"annotated bool"`, so the filter should see it already

drin: but no e2e test on a renamed column because a rename has to happen in a model body, we cannot rename with schemas.

bigluck (approved):
well done! thanks!

